PR #4, opened 2020-03-10
BOTMETA

gundalow:
##### SUMMARY
Unsure if we will need ansibulbot on this repo, though here is BOTMETA incase we do.


jillr:
This rebase got pretty bad, have opened #6 from a clean branch to add this file.

PR #25, opened 2020-04-03
Add hostname options concatenation in aws_ec2 inventory plugin

j8r:
Reborn of https://github.com/ansible/ansible/pull/58052.
Fixes https://github.com/ansible/ansible/issues/55911.
##### SUMMARY
Adds hostname options concatenation in the aws_ec2 inventory plugin. 

This change adds support for list of options, that will be concatenated with an by default an underscore to create the inventory name.

Example:
```yml
hostnames:
  - name: 'private-ip-address'
    prefix: 'tag:Name'
```
Result: `myinstance_10.10.10.10`

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ansible.inventory.plugins.aws_ec2

##### ADDITIONAL INFORMATION
[Amazon Auto Scaling Groups](https://docs.aws.amazon.com/autoscaling/ec2/userguide/AutoScalingGroup.html) allows to have a group of one or multiple instances, that scales depending of policies.
However, they all share the same tags and dns name. If the ASG includes several instances, only 1 will be targeted when using this options. IPs and Instance IDs are unique, but using them isn't user-friendly.

In order to still use the tag:Name, an unique identifier like an IP or an Instance ID has to be appended.

### DIFF
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -512,9 +512,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 if 'name' not in preference:
                     raise AnsibleError("A 'name' key must be defined in a hostnames dictionary.")
                 hostname = self._get_hostname(instance, [preference["name"]])
+                hostname_from_prefix = self._get_hostname(instance, [preference["prefix"]])
                 separator = preference.get("separator", "_")
-                if hostname and 'prefix' in preference:
-                    hostname_from_prefix = self._get_hostname(instance, [preference["prefix"]]) or ''
+                if hostname and hostname_from_prefix and 'prefix' in preference:
                     hostname = hostname_from_prefix + separator + hostname
             elif preference.startswith('tag:'):
                 if not re.match(r'^tag:\S+=\S+$', preference):
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -514,9 +514,10 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 hostname = self._get_hostname(instance, [preference["name"]])
                 separator = preference.get("separator", "_")
                 if hostname and 'prefix' in preference:
-                    hostname = self._get_hostname(instance, [preference["prefix"]]) + separator + hostname
-            elif 'tag' in preference:
-                if not preference.startswith('tag:'):
+                    hostname_from_prefix = self._get_hostname(instance, [preference["prefix"]]) or ''
+                    hostname = hostname_from_prefix + separator + hostname
+            elif preference.startswith('tag:'):
+                if not re.match(r'^tag:\S+=\S+$', preference):
                     raise AnsibleError("To name a host by tags name_value, use 'tag:name=value'.")
                 hostname = self._get_tag_hostname(preference, instance)
             else:
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -57,6 +57,7 @@ DOCUMENTATION = '''
               - Spot instances may be persistent and instances may have associated events.
           type: bool
           default: False
+          version_added: '1.2.2'
         strict_permissions:
           description:
               - By default if a 403 (Forbidden) error code is encountered this plugin will fail.
@@ -76,6 +77,7 @@ DOCUMENTATION = '''
               which group names end up being used as.
           type: bool
           default: False
+          version_added: '1.2.2'
 '''
 
 EXAMPLES = '''
@@ -105,10 +107,13 @@ strict_permissions: False
 # Note: I(hostnames) sets the inventory_hostname. To modify ansible_host without modifying
 # inventory_hostname use compose (see example below).
 hostnames:
-  - tag:Name=Tag1,Name=Tag2  # Return specific hosts only
+  - tag:Name=Tag1,Name=Tag2         # Return specific hosts only
   - tag:CustomDNSName
   - dns-name
-  - private-ip-address
+  - name: 'tag:Name=Tag1,Name=Tag2'
+  - name: 'private-ip-address'
+    separator: '_'
+    prefix: 'tag:Name'
 
 # Example using constructed features to create groups and set ansible_host
 plugin: aws_ec2
@@ -505,7 +510,14 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
         hostname = None
         for preference in hostnames:
-            if 'tag' in preference:
+            if isinstance(preference, dict):
+                if 'name' not in preference:
+                    raise AnsibleError("A 'name' key must be defined in a hostnames dictionary.")
+                hostname = self._get_hostname(instance, [preference["name"]])
+                separator = preference.get("separator", "_")
+                if 'prefix' in preference:
+                    hostname = self._get_hostname(instance, [preference["prefix"]]) + separator + hostname
+            elif 'tag' in preference:
                 if not preference.startswith('tag:'):
                     raise AnsibleError("To name a host by tags name_value, use 'tag:name=value'.")
                 hostname = self._get_tag_hostname(preference, instance)
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -517,8 +517,6 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 if hostname and hostname_from_prefix and 'prefix' in preference:
                     hostname = hostname_from_prefix + separator + hostname
             elif preference.startswith('tag:'):
-                if not re.match(r'^tag:\S+=\S+$', preference):
-                    raise AnsibleError("To name a host by tags name_value, use 'tag:name=value'.")
                 hostname = self._get_tag_hostname(preference, instance)
             else:
                 hostname = self._get_boto_attr_chain(preference, instance)
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -57,7 +57,6 @@ DOCUMENTATION = '''
               - Spot instances may be persistent and instances may have associated events.
           type: bool
           default: False
-          version_added: '1.2.2'
         strict_permissions:
           description:
               - By default if a 403 (Forbidden) error code is encountered this plugin will fail.
@@ -77,7 +76,6 @@ DOCUMENTATION = '''
               which group names end up being used as.
           type: bool
           default: False
-          version_added: '1.2.2'
 '''
 
 EXAMPLES = '''
@@ -107,7 +105,7 @@ strict_permissions: False
 # Note: I(hostnames) sets the inventory_hostname. To modify ansible_host without modifying
 # inventory_hostname use compose (see example below).
 hostnames:
-  - tag:Name=Tag1,Name=Tag2         # Return specific hosts only
+  - tag:Name=Tag1,Name=Tag2  # Return specific hosts only
   - tag:CustomDNSName
   - dns-name
   - name: 'tag:Name=Tag1,Name=Tag2'
@@ -515,7 +513,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                     raise AnsibleError("A 'name' key must be defined in a hostnames dictionary.")
                 hostname = self._get_hostname(instance, [preference["name"]])
                 separator = preference.get("separator", "_")
-                if 'prefix' in preference:
+                if hostname and 'prefix' in preference:
                     hostname = self._get_hostname(instance, [preference["prefix"]]) + separator + hostname
             elif 'tag' in preference:
                 if not preference.startswith('tag:'):
--- a/tests/integration/targets/inventory_aws_ec2/playbooks/test_populating_inventory_with_concatenation.yml
+++ b/tests/integration/targets/inventory_aws_ec2/playbooks/test_populating_inventory_with_concatenation.yml
@@ -36,12 +36,12 @@
 
         - name: register the current hostname
           set_fact:
-            hostname: "{{ resource_prefix | replace('-', '_') }}_value"
+            expected_hostname: "value_{{ resource_prefix }}"
 
-        - name: "assert the hostname  is properly formed, '{{ hostname }}' should equal '{{ inventory_hostname }}'"
+        - name: "Ensure we've got a hostvars entry for the new host"
           assert:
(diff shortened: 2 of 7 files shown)


Dmitry1987:
How the aws_ec2 plugin was even useful before this PR?  Like, isn't half of the use cases on AWS are autoscaling groups?  this is a core base feature, please merge-merge-merge-merge-merge someone!!!

j8r:
The code was already reviewed by @s-hertel .
Remains another pass.
Would someone could find the time for a little review, maybe @jillr ?
Thank you :)

jillr:
Thanks for carrying this over @j8r - I will take a look at this PR this week.  Would you be able to add a changelog fragment to the `changelogs/fragments` directory?  Some info about how changelogs are handled in collections can be found in the [antsibull-changelog](https://github.com/ansible-community/antsibull-changelog/blob/main/docs/changelogs.rst) project.  You may need to rebase this PR against the latest upstream, the default branch for that is now named `main`.

j8r:
I have run `antsibull-changelog generate --refresh --reload-plugins` but all files in `changelogs/fragments` have been removed.

jillr:
@j8r if you're not trying to generate a final changelog for a release you won't need to run that command.  We just need a fragment that matches the format the tool will expect so that it will be incorporated at the next release of the collection.

plugins/inventory/aws_ec2.py
```diff
@@ -57,6 +57,7 @@
               - Spot instances may be persistent and instances may have associated events.
           type: bool
           default: False
+          version_added: '2.8'
```
jillr, line 60:
```suggestion
          version_added: '1.1.0'
```
This will be the next release version of the collection.

plugins/inventory/aws_ec2.py
```diff
@@ -76,6 +77,7 @@
               which group names end up being used as.
           type: bool
           default: False
+          version_added: '2.8'
```
jillr, line 80:
```suggestion
          version_added: '1.1.0'
```

plugins/inventory/aws_ec2.py
```diff
@@ -505,7 +510,14 @@ def _get_hostname(self, instance, hostnames):
 
         hostname = None
         for preference in hostnames:
-            if 'tag' in preference:
+            if isinstance(preference, dict):
+                if 'name' not in preference:
+                    raise AnsibleError("A 'name' key must be defined in a hostnames dictionnary.")
```
jillr, line 515:
```suggestion
                    raise AnsibleError("A 'name' key must be defined in a hostnames dictionary.")
```

jillr:
@s-hertel you had previously reviewing this PR in ansible/ansible, are you happy with the current status?

j8r:
Ok, I will add a fragment, then lint it. What name should it take, I see there is a kind of ID at first (maybe issue ID)?

j8r:
Thanks for the review. I suppose I have to also create a new playbook along with the template in `tests/integration/targets/inventory_aws_ec2`, and a command to execute it in `runme.sh`?

jillr:
Generally the PR number and a description of the change. So you might use something like 25-aws_ec2-hostname-options-concatenation.yaml

And yes, a new template and playbook in runme.sh (you should be able to copy one of the other playbooks as something like test_populating_inventory_with_concatenation.yml and modify it to have the correct assertions).

Thanks!

j8r:
I succeeded to have all the glue code around the test working, but I'm a bit stuck how to have an assert integration test for this, the hostname is `127.0.0.1` in the first place.

j8r:
Any update?

noahjd:
Any update on this?

goneri:
Hi! I assign this to myself. We would like to land this quickly.

j8r:
Ho, thanks @goneri . Essentially, I struggle with the integration test.
I rebase the branch to make ready it for review.

plugins/inventory/aws_ec2.py
```diff
@@ -505,7 +510,14 @@ def _get_hostname(self, instance, hostnames):
...
         hostname = None
         for preference in hostnames:
-            if 'tag' in preference:
+            if isinstance(preference, dict):
+                if 'name' not in preference:
+                    raise AnsibleError("A 'name' key must be defined in a hostnames dictionary.")
+                hostname = self._get_hostname(instance, [preference["name"]])
+                separator = preference.get("separator", "_")
+                if 'prefix' in preference:
+                    hostname = self._get_hostname(instance, [preference["prefix"]]) + separator + hostname
```
goneri, line 519: If `_get_hostname` return `None` (e.g: tag is missing on a VM), I get:
```
unsupported operand
type(s) for +: 'NoneType' and 'AnsibleUnicode'
```
So it would be better to do it in two steps:

```python
if 'prefix' in preference and self._get_hostname(instance, [preference["prefix"]]):
    hostname = self._get_hostname(instance, [preference["prefix"]]) + separator + hostname
```

goneri (changes requested):
Thank you for the PR, this is a couple of things to improve before we can merge it.

plugins/inventory/aws_ec2.py
j8r: Done, added a leading `hostname and` to the condition

plugins/inventory/aws_ec2.py
```diff
@@ -505,7 +508,14 @@ def _get_hostname(self, instance, hostnames):
...
         for preference in hostnames:
-            if 'tag' in preference:
+            if isinstance(preference, dict):
+                if 'name' not in preference:
+                    raise AnsibleError("A 'name' key must be defined in a hostnames dictionary.")
+                hostname = self._get_hostname(instance, [preference["name"]])
+                separator = preference.get("separator", "_")
+                if hostname and 'prefix' in preference:
+                    hostname = self._get_hostname(instance, [preference["prefix"]]) + separator + hostname
+            elif 'tag' in preference:
```
goneri, line 518: This sounds wrong to me. This block should indented one level more.

goneri: I suggest something like that instead:
```patch
diff --git a/plugins/inventory/aws_ec2.py b/plugins/inventory/aws_ec2.py
index 29fa9a1..12d0e52 100644
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -515,8 +515,8 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 separator = preference.get("separator", "_")
                 if hostname and 'prefix' in preference:
                     hostname = self._get_hostname(instance, [preference["prefix"]]) + separator + hostname
-            elif 'tag' in preference:
-                if not preference.startswith('tag:'):
+            if preference.startswith('tag:'):
+                if not re.match(r'^tag:\S+=\S+$', preference):
                     raise AnsibleError("To name a host by tags name_value, use 'tag:name=value'.")
                 hostname = self._get_tag_hostname(preference, instance)
             else:
```

plugins/inventory/aws_ec2.py
goneri: This does not resolve my problem, `self._get_hostname()` can still return `None` if there is no match. I use this patch: 

```
diff --git a/plugins/inventory/aws_ec2.py b/plugins/inventory/aws_ec2.py
index 29fa9a1..2d6568f 100644
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -514,10 +514,9 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 hostname = self._get_hostname(instance, [preference["name"]])
                 separator = preference.get("separator", "_")
                 if hostname and 'prefix' in preference:
-                    hostname = self._get_hostname(instance, [preference["prefix"]]) + separator + hostname
-            elif 'tag' in preference:
-                if not preference.startswith('tag:'):
-                    raise AnsibleError("To name a host by tags name_value, use 'tag:name=value'.")
+                    hostname_from_prefix = self._get_hostname(instance, [preference["prefix"]]) or ''
+                    hostname = hostname_from_prefix + separator + hostname
+            elif preference.startswith('tag:'):
                 hostname = self._get_tag_hostname(preference, instance)
             else:
                 hostname = self._get_boto_attr_chain(preference, instance)
```

With this configuration:

```yaml
hostnames:                                                                                    
  - name: 'tag:Name=aws:cloudformation:stack-id'                                              
  - name: 'private-ip-address'                                                                
    separator: '_'                                                                            
    prefix: 'tag:Name'
```

I get:

- _10.0.0.43 <---- There is no matching tag for this one, so the prefix is ''
- _10.0.234.96 <---- the same
- ansible-operator-rhel-to-rhel-responder-instance_192.168.0.10
- cvpn-initiator-instance_10.0.0.10
            "cvpn-responder-instance_192.168.0.10"

My question at this stage, should we skip these two first instances, or keep them in the list?

goneri:
Thanks for the update, I will take a look that the test-suite. In the meantime, you can ping me on IRC if you need help or have question: I'm goneri on #ansible-aws or #ansible-cloud.

plugins/inventory/aws_ec2.py
j8r: Sometimes, at the very early stage of initialization, all instance tags may not have been set.
We have solved this on our case by having a wait loop in the cloud-init.

I don't know if it is useful to have such inventory names with no prefix, though.
I propose to skip them, and if someone has a use-case, we could in the future add them.

plugins/inventory/aws_ec2.py
```diff
@@ -505,8 +508,16 @@ def _get_hostname(self, instance, hostnames):
...
+            if isinstance(preference, dict):
+                if 'name' not in preference:
+                    raise AnsibleError("A 'name' key must be defined in a hostnames dictionary.")
+                hostname = self._get_hostname(instance, [preference["name"]])
+                hostname_from_prefix = self._get_hostname(instance, [preference["prefix"]])
+                separator = preference.get("separator", "_")
+                if hostname and hostname_from_prefix and 'prefix' in preference:
+                    hostname = hostname_from_prefix + separator + hostname
+            elif preference.startswith('tag:'):
+                if not re.match(r'^tag:\S+=\S+$', preference):
```
goneri, line 520: The tests are failing because of this test since `tag:Name` is not accepted anymore. I think we can just just drop these two lines.

j8r: Dropping this this `elif` condition block? What do you mean by `tag:Name` no longer supported?

goneri: Just the line 520 and 521. The regex does not match `tag:Name`.

goneri: For clarity, this is what I mean:

```patch
diff --git a/plugins/inventory/aws_ec2.py b/plugins/inventory/aws_ec2.py
index c382f34..f217e79 100644
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -517,8 +517,6 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
                 if hostname and hostname_from_prefix and 'prefix' in preference:
                     hostname = hostname_from_prefix + separator + hostname
             elif preference.startswith('tag:'):
-                if not re.match(r'^tag:\S+=\S+$', preference):
-                    raise AnsibleError("To name a host by tags name_value, use 'tag:name=value'.")
                 hostname = self._get_tag_hostname(preference, instance)
             else:
                 hostname = self._get_boto_attr_chain(preference, instance)
```

j8r: Thanks. Note also that there is a feature in GitHub to select multiple lines and then add a comment for them.

goneri: :-) Thanks for the suggestion.

tests/integration/targets/inventory_aws_ec2/playbooks/test_populating_inventory_with_concatenation.yml
```diff
@@ -0,0 +1,61 @@
...
+        - meta: refresh_inventory
+
+        - name: register the current hostname
+          set_fact:
+            hostname: "{{ resource_prefix | replace('-', '_') }}_value"
+
+        - name: "assert the hostname  is properly formed, '{{ hostname }}' should equal '{{ inventory_hostname }}'"
+          assert:
+            that:
+              - hostname == inventory_hostname
```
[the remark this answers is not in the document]
goneri: This is our last CI error :-). `inventory_hostname` remains 127.0.0.1. I think we can just check in `hostvars` if we've got the new host.

```patch
diff --git a/tests/integration/targets/inventory_aws_ec2/playbooks/test_populating_inventory_with_concatenation.yml b/tests/integration/targets/inventory_aws_ec2/playbooks/test_populating_inventory_with_concatenation.yml
index 20b45f0..39d28d1 100644
--- a/tests/integration/targets/inventory_aws_ec2/playbooks/test_populating_inventory_with_concatenation.yml
+++ b/tests/integration/targets/inventory_aws_ec2/playbooks/test_populating_inventory_with_concatenation.yml
@@ -36,12 +36,12 @@
 
         - name: register the current hostname
           set_fact:
-            hostname: "{{ resource_prefix | replace('-', '_') }}_value"
+            expected_hostname: "{{ resource_prefix | replace('-', '_') }}_value"
 
-        - name: "assert the hostname  is properly formed, '{{ hostname }}' should equal '{{ inventory_hostname }}'"
+        - name: "Ensure we've got a hostvars entry for the new host"
           assert:
             that:
-              - hostname == inventory_hostname
+              - expected_hostname in hostvars
 
       always:
```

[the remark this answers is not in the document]
goneri: The patch above is broken. I'm working on a better solution.

goneri:
With this patch, I can run the test-suite correctly: https://github.com/ansible-collections/amazon.aws/commit/c5a85be54c0101694e58f2870c4b515e21bf55e5.patch

goneri (approved):

goneri (approved):

jillr (approved):
LGTM, thanks @j8r!

vivanov-dp:
Hey, this looks great and will be really useful to us once it gets released, but it looks like it only allows 2 options to be concatenated, and judging by a comment here: https://github.com/ansible/ansible/issues/55911#issuecomment-653209122 this might not be enough for some use-cases.

Please consider adding something like what I did in #231 to it in the future.

goneri:
@j8r, I feel like @vivanov-dp suggestion would actually be a better long-term solution. Do you think you can build up a solution together before the next release? cc: @jillr, @tremble

Issue #39, opened 2020-04-27
Tags need to be added to galaxy.yml


JPrause:
Tags are needed for Collection v1.0
The current value for tags is null

PR #40, opened 2020-04-28
Adding labels to the tags section

JPrause:
Adding tags to resolve https://github.com/ansible-collections/amazon.aws/issues/39

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -7,7 +7,7 @@ authors:
 description: null
 license: GPL-3.0-or-later
 license_file: COPYING
-tags: [amazon, aws, cloud, networking]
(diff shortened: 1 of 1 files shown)


galaxy.yml
```diff
@@ -7,7 +7,7 @@ authors:
 description: null
 license: GPL-3.0-or-later
 license_file: COPYING
-tags: null
+tags: [amazon, aws, cloud, networking]
```
jillr, line 10: I'm not sure networking makes sense here; can you help me understand this tag's inclusion?

JPrause: Thanks @jillr for reviewing.
I was looking at another example when I typed that.
I'll submit an update to remove it.

PR #60, opened 2020-05-28
Update action groups

s-hertel:
##### SUMMARY
The format expected by https://github.com/ansible/ansible/pull/67291 changed, so just updating these here.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
action groups

PR #69, opened 2020-06-10
Specify test requirements for collection

jillr:
##### SUMMARY
default test container has been slimmed down; we need to specify all requirements now

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/integration/requirements.txt

### DIFF
--- a/tests/integration/requirements.txt
+++ b/tests/integration/requirements.txt
@@ -1,0 +1,2 @@
+# netaddr is needed for ansible.netcommon.ipv6
+netaddr


jillr:
Right now we are in a spot where #60 depends on this for the `netaddr` dependency, and this depends on 60 for the new module_defaults/runtime.yml definition.

This test: https://app.shippable.com/github/ansible-collections/amazon.aws/runs/282/summary/console
Is a successful experiment of rebasing this PR on top of 60 and running ci_complete.  In light of that, and since this PR is the smaller/more trivial of the two patches, I'm going to re-push this branch after rebasing back onto master, let CI end up with whatever result it's going to end up with, and merge on the basis of this test run.

PR #73, opened 2020-06-17
Use git tags for version number instead of galaxy.yml

jillr:
##### SUMMARY
Use git tags for version number instead of galaxy.yml

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
galaxy.yml


jillr:
We can't actually remove this with shippable testing.  Zuul will override this value so we can leave it for now.

PR #80, opened 2020-06-18
Don't run code coverage analyze on 2.9

jillr:
##### SUMMARY

This wasn't added until 2.10; #67141
See failures in https://app.shippable.com/github/ansible-collections/amazon.aws/runs/333/8/console

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/utils/shippable/shippable.sh

### DIFF
--- a/tests/utils/shippable/shippable.sh
+++ b/tests/utils/shippable/shippable.sh
@@ -118,8 +118,10 @@ function cleanup
             ansible-test coverage xml --color --requirements --group-by command --group-by version ${stub:+"$stub"}
             cp -a tests/output/reports/coverage=*.xml "$SHIPPABLE_RESULT_DIR/codecoverage/"
 
-            # analyze and capture code coverage aggregated by integration test target
-            ansible-test coverage analyze targets generate -v "$SHIPPABLE_RESULT_DIR/testresults/coverage-analyze-targets.json"
+            # analyze and capture code coverage aggregated by integration test target if not on 2.9
+            if [ -n "${ANSIBLE_BASE_REV:-}" ] || [ "${ANSIBLE_BASE_REV:-}" != "2.9" ]; then
+                ansible-test coverage analyze targets generate -v "$SHIPPABLE_RESULT_DIR/testresults/coverage-analyze-targets.json"
+            fi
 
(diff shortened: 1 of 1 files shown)


tests/utils/shippable/shippable.sh
```diff
@@ -118,8 +118,10 @@ function cleanup
             ansible-test coverage xml --color --requirements --group-by command --group-by version ${stub:+"$stub"}
             cp -a tests/output/reports/coverage=*.xml "$SHIPPABLE_RESULT_DIR/codecoverage/"
 
-            # analyze and capture code coverage aggregated by integration test target
-            ansible-test coverage analyze targets generate -v "$SHIPPABLE_RESULT_DIR/testresults/coverage-analyze-targets.json"
+            # analyze and capture code coverage aggregated by integration test target if not on 2.9
+            if [ -n "${ANSIBLE_BASE_REV:-}" ] || [ "${ANSIBLE_BASE_REV:-}" != "2.9" ]; then
```
goneri, line 122: It should just be:
`if [ "${ANSIBLE_BASE_REV:-}" != "2.9" ]; then`

Otherwise, `ANSIBLE_BASE_REV` is not empty, and consequently the left block is true.

jillr: The default is to be empty, today it might be set to 2.9 but in the very near future it could also be 2.10.  So the goal is 'variable is empty OR is not 2.9'

goneri: You want to run `ansible-test coverage analyze targets generate` if `ANSIBLE_BASE_REF` is empty or if it's value is `2.9`. Right?

jillr: No, I want to run it if the value is empty or not 2.9 - the command wasn't added until 2.10.

jillr: (Empty ends up being devel)

goneri: Oh! thanks for the clarification :-)

jillr: I'll update the comment to make it more clear before I merge, thanks for reviewing @goneri!

PR #83, opened 2020-06-19
Add shippable matrix jobs for 2.9, 2.10

jillr:
##### SUMMARY
Run all supported Ansible revisions in test matrix

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
shippable.sh
shippable.yml


shippable.yml
```diff
@@ -9,24 +9,52 @@ matrix:
...
+    - env: T=units/3.7/1 A_REV=2.10
+    - env: T=units/3.8/1 A_REV=2.10
+    - env: T=units/3.9/1 A_REV=2.10
 
     - env: T=aws/2.7/1
     - env: T=aws/3.7/1
+    - env: T=aws/2.7/1 A_REV=2.9
+    - env: T=aws/2.7/1 A_REV=2.9
+    - env: T=aws/3.7/1 A_REV=2.10
+    - env: T=aws/3.7/1 A_REV=2.10
```
mattclay, line 36: Mixed up Python versions here:

```suggestion
    - env: T=aws/2.7/1 A_REV=2.9
    - env: T=aws/3.7/1 A_REV=2.9
    - env: T=aws/2.7/1 A_REV=2.10
    - env: T=aws/3.7/1 A_REV=2.10
```

shippable.yml
```diff
@@ -9,24 +9,52 @@ matrix:
...
     - env: T=units/2.7/1
     - env: T=units/3.6/1
     - env: T=units/3.7/1
     - env: T=units/3.8/1
     - env: T=units/3.9/1
+    - env: T=units/2.7/1 A_REV=2.9
+    - env: T=units/3.6/1 A_REV=2.9
+    - env: T=units/3.7/1 A_REV=2.9
+    - env: T=units/3.8/1 A_REV=2.9
+    - env: T=units/3.9/1 A_REV=2.9
```
mattclay, line 24: Ansible 2.9 does not support Python 3.9:

```suggestion
    - env: T=units/3.9/1 A_REV=2.9
```

tests/utils/shippable/shippable.sh
```diff
@@ -75,8 +75,8 @@ set -ux
 
 pip install setuptools==44.1.0
 
-if [ -n "${ANSIBLE_BASE_REV:-}" ]; then
-    pip install "ansible~=${ANSIBLE_BASE_REV}"
+if [ -n "${A_REV:-}" ]; then
+    pip install "ansible~=${A_REV}"
```
mattclay, line 79: Unfortunately this needs to be a bit more complex since you probably want to use `ansible` for 2.9 and `ansible-base` for 2.10 and later. Also, since you want to test a pre-release, `~=` doesn't work, even with `--pre`, so you'll need to pick a specific version. Either that or specify a branch name instead of a version number.

Here's one possibility using version numbers:

```suggestion
if [[ "${A_REV:-}" == "2.9" ]]; then
    install_package="ansible~=2.9"
elif [[ "${A_REV:-}" == 2.10 ]]; then
    install_package="ansible-base==2.10.0b1"
fi

if [ -n "${install_package:-}" ]; then
    pip install "${install_package}"
```

If you wanted to specify a branch name you could do something like `A_BR=2.10` and have the script use `"stable-${A_BR}"` instead of `"devel"` if it is set (instead of using `A_REV`).

shippable.yml
```diff
@@ -9,24 +9,51 @@ matrix:
     - env: T=none
   include:
     - env: T=sanity/1
+    - env: T=sanity/1 A_REV=2.9
+    - env: T=sanity/1 A_REV=2.10.0a1
```
mattclay, line 13: Are you sure you want to test against ansible 2.10 instead of ansible-base 2.10? Doesn't this collection already exist in ansible 2.10?

PR #87, opened 2020-06-24
Cleanup sanity-ignores

jillr:
##### SUMMARY
Cleanup sanity-ignores

Mainly add element types to lists, future imports/metatype to tests.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
sanity


tests/unit/mock/path.py
```diff
@@ -1,5 +1,8 @@
 from ansible_collections.amazon.aws.tests.unit.compat.mock import MagicMock
 from ansible.utils.path import unfrackpath
 
+from __future__ import (absolute_import, division, print_function)
```
Akasurde, line 4: I think this should go above imports.

PR #95, opened 2020-07-08
Enable all sanity tests

felixfontein:
##### SUMMARY
Right now, the `pylint`, `validate-modules` and `ansible-doc` sanity tests are not run in CI. This removes the expectation from tests/utils/shippable/sanity.sh that there are five sanity check groups so that all tests are run.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
CI

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,5 +1,5 @@
 ---
-requires_ansible: '>=2.9,<2.11'
+requires_ansible: '>=2.9'
 action_groups:
   aws:
   - aws_s3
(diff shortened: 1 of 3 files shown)


felixfontein:
The remaining failed sanity test (https://app.shippable.com/github/ansible-collections/amazon.aws/runs/429/1/tests) is because `devel` branch is now version 2.11, and this collection claims it requires Ansible < 2.11. I'm not sure how this is best fixed.

PR #176, opened 2020-10-21
Clean up test requirements

jillr:
##### SUMMARY
We have unit test requirements from ansible/ansible that aren't needed for aws

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/unit/requirements.txt


tremble (approved):
Wow that's a lot of cruft we pulled over.  No big surprise I guess.

PR #184, opened 2020-10-29
feat: Add helper function scrub_none_parameters

wimnat:
##### SUMMARY
Adding a helper function to scrub dicts of keys that have a value of None.

This is going to be a common scenario due to the `elements='dict'` option being added in module options.

See https://github.com/ansible-collections/community.aws/issues/251 for more detail.

I have added in core.py rather than ec2.py because it should be useful for all amazon modules

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module_utils core

##### ADDITIONAL INFORMATION
None


plugins/module_utils/core.py
```diff
@@ -357,3 +357,27 @@ def get_boto3_client_method_parameters(client, method_name, required=False):
...
+    Credit: https://medium.com/better-programming/how-to-remove-null-none-values-from-a-dictionary-in-python-1bedf1aab5e4
+
+    :param parameters: parameter dict
+    :return: parameter dict with all keys = None removed
+    """
+
+    clean_parameters = {}
+
+    for k, v in parameters.items():
+        if isinstance(v, dict):
```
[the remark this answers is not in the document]
wimnat: See line 380.

Do you think we should do something different?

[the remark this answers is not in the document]
markuman: Ah. I misread it.

markuman:
LGTM, but CI is failing.

plugins/module_utils/core.py
```diff
@@ -357,3 +357,27 @@ def get_boto3_client_method_parameters(client, method_name, required=False):
...
+            nested = scrub_none_parameters(v)
+            if len(nested.keys()) > 0:
```
felixfontein, line 378: The more Pythonic way would be:
```suggestion
            if nested:
```
In any case: is this needed? I don't know the AWS APIs, but maybe there are cases where an empty dictionary is treated differently from a not specified one.

tests/unit/module_utils/core/test_core.py
```diff
@@ -0,0 +1,32 @@
...
+    (dict(param1='something', param2=None, param3=None, param4='something_else'),
+     dict(param1='something', param4='something_else')
+     ),
+    (dict(param1=None, param2=None),
+     dict()
+     )
+]
+
+
+@pytest.mark.parametrize("input_params, output_params", scrub_none_test_data)
```
felixfontein, line 29: You don't have any tests for recursion, and especially not the condition that empty sub-dicts are also removed.

plugins/module_utils/core.py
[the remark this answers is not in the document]
felixfontein: Also, similar to empty sub-dictionaries, should empty lists be removed? (I don't know whether it's a good idea. That's something that should be checked first by someone who knows the APIs better.)

Maybe the two behaviors need to be configurable.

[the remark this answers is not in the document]
tremble: The main use case I see is purging parameters that are ``None`` from the args as parse by Ansible(AWS)Module.  

The worst offenders being: ``(type=list, options=...)`` and ``(type=dict, options=...)``.  You've covered the second option, but the first is currently left as an exercise for the reader.  What might be nice is to support optionally descending into lists, dropping None, and None from dicts inside the list...

plugins/module_utils/core.py
wimnat: Even if not in the AWS APIs I know that modules themselves depend on this behaviour.  For example, an empty dict may remove all tags where as an unspecified dict will leave tags alone.

tremble (changes requested):
Thanks for taking the time on this, in the longer term this feels like something which belongs in Ansible itself, but an initial implementation that proves its use here in amazon.aws is probably the right way to go.

I like the idea of making the recursion into lists and removal of empty iterables configurable behaviour, I suspect that both behaviours will be needed 'somewhere'

Couple of minor niggles while the other discussions are ongoing:
- please add a changelog fragment.
- please move the test to tests/unit/module_utils/core/test_scrub_none_parameters.py

tests/unit/module_utils/core/test_core.py
wimnat: I've added some additional tests to cover empty sub-dicts, sub-dicts as None and empty lists

plugins/module_utils/core.py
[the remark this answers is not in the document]
wimnat: Thanks for the feedback.

@tremble I agree with your main use case.  I have altered the code to only discard params set to None.  Empty dicts or lists will be kept because they are likely wanted.  I know modules for example that will take an empty dict for the `tags` param to remove all tags.

The code will also drop lists set as None so will handle (type=list, options=...).

tests/unit/module_utils/core/test_scrub_none_parameters.py
```diff
@@ -0,0 +1,47 @@
+from plugins.module_utils.core import scrub_none_parameters
+import pytest
+
+scrub_none_test_data = [
+    (dict(),
+     dict()
+     ),
+    (dict(param1='something'),
+     dict(param1='something')
+     ),
```
tremble, line 10:
```suggestion
    (dict(param1='something'),
     dict(param1='something')
     ),
    (dict(param1=False),
     dict(param1=False)
     ),
```

tests/unit/module_utils/core/test_scrub_none_parameters.py
```diff
@@ -0,0 +1,47 @@
...
+    (dict(param1=dict(sub_param1='something', sub_param2=None), param2=None, param3=None, param4='something_else'),
+     dict(param1=dict(sub_param1='something'), param4='something_else')
+     ),
+    (dict(param1=dict(sub_param1='something', sub_param2=dict(sub_sub_param1='another_thing')), param2=None, param3=None, param4='something_else'),
+     dict(param1=dict(sub_param1='something', sub_param2=dict(sub_sub_param1='another_thing')), param4='something_else')
+     ),
+    (dict(param1=dict(sub_param1='something', sub_param2=dict()), param2=None, param3=None, param4='something_else'),
+     dict(param1=dict(sub_param1='something', sub_param2=dict()), param4='something_else')
+     ),
```
tremble, line 34:
```suggestion
    (dict(param1=dict(sub_param1='something', sub_param2=dict()), param2=None, param3=None, param4='something_else'),
     dict(param1=dict(sub_param1='something', sub_param2=dict()), param4='something_else')
     ),
    (dict(param1=dict(sub_param1='something', sub_param2=False), param2=None, param3=None, param4='something_else'),
     dict(param1=dict(sub_param1='something', sub_param2=False), param4='something_else')
     ),     
```

tests/unit/module_utils/core/test_scrub_none_parameters.py
```diff
@@ -0,0 +1,47 @@
+from plugins.module_utils.core import scrub_none_parameters
```
tremble, line 1:
```suggestion
from __future__ import (absolute_import, division, print_function)
__metaclass__ = type

from ansible_collections.amazon.aws.plugins.module_utils.core import scrub_none_parameters
```

tremble (changes requested):
Feature-wise I think this is a good first pass and should let you get the fix in for ansible-collections/community.aws#251

There's a couple of extra tests I'd like to see, just because they'd catch dumb mistakes I've seen over the years (including made by me).

I'm torn WRT whether we should descend into lists too, so I'd like @jillr's take on it.  I have a very soft preference for purging lists of None and descending into any of the dicts inside the list, but it's soft enough that I'd follow up with a second PR rather than blocking this PR.

tests/unit/module_utils/core/test_scrub_none_parameters.py
wimnat: Just for my understanding, why do we need this change?

tremble: 'plugins.module_utils' vs 'ansible_collections.amazon.aws.plugins.module_utils' this is mostly an artifact of the collections migration: collections now get their own namespace.

the future / metaclass boiler plate is the same as ever it's just ensuring python 2 and 3 behave a little bit more consistently.

tests/unit/module_utils/core/test_scrub_none_parameters.py
```diff
@@ -0,0 +1,56 @@
...
+     ),
+    (dict(param1=dict(sub_param1='something', sub_param2=dict(sub_sub_param1='another_thing')), param2=None, param3=None, param4='something_else'),
+     dict(param1=dict(sub_param1='something', sub_param2=dict(sub_sub_param1='another_thing')), param4='something_else')
+     ),
+    (dict(param1=dict(sub_param1='something', sub_param2=dict()), param2=None, param3=None, param4='something_else'),
+     dict(param1=dict(sub_param1='something', sub_param2=dict()), param4='something_else')
+     ),
+    (dict(param1=dict(sub_param1='something', sub_param2=False), param2=None, param3=None, param4='something_else'),
+     dict(param1=dict(sub_param1='something', sub_param2=False), param4='something_else')
+     ),
```
tremble, line 43:
```suggestion
     ),
```

tremble (approved):

wimnat:
@felixfontein are you happy with this or anything else required?

changelogs/fragments/184-scrub-none-params.yaml
```diff
@@ -0,0 +1,4 @@
+---
+minor_changes:
+  - module_utils/core - Add helper function scrub_none_parameters to remove params set to None. Enabler
+    to fix issue (https://github.com/ansible-collections/community.aws/issues/251)
```
felixfontein, line 4:
```suggestion
  - module_utils/core - add helper function ``scrub_none_parameters`` to remove params set to ``None`` (https://github.com/ansible-collections/community.aws/issues/251).
```
I removed the `Enabler to fix issue` part because IMO it is hard to understand, and not really needed because the link indicates this is related to the issue.

felixfontein (approved):

markuman (approved):

tremble:
LGTM

Good to merge once the tests pass, original failures weren't related to this change (ENI work in progress - #180, Access looks to be AWS getting grumpy due to the number of failed auth events)

felixfontein:
@wimnat thanks for implementing this!
@markuman @tremble thanks for reviewing this!

PR #191, opened 2020-11-05
Add assume role functionality to aws_rds

hamaker:
I've basically copied the relevant functions from `aws_ec2.py`, there's nothing really new in here. 
Probably this shared functionality would be better to put in a shared module, as I've mostly just duplicated code. But still I think this is a useful first step.

##### SUMMARY
Carry over the asume role functionality from aws_ec2.py

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
aws_rds

##### ADDITIONAL INFORMATION
As I said above, I think there's some shared functionality between the EC2 and RDS inventory plugins that could be extracted into a shared module. I didn't want to make too many changes in the code, and as there was already some duplication between the two plugins, so I basically followed what was already there. i'd be happy to refactor this a bit, please let me know if that would be appreciated.

### DIFF
--- a/plugins/inventory/aws_rds.py
+++ b/plugins/inventory/aws_rds.py
@@ -36,6 +36,9 @@ DOCUMENTATION = '''
           default:
               - creating
               - available
+        iam_role_arn:
+          description: The ARN of the IAM role to assume to perform the inventory lookup. You should still provide
+              AWS credentials with enough privilege to perform the AssumeRole action.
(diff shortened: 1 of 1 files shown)


hamaker:
I forgot to mention that this would resolve issue #30

jillr (approved):
LGTM, tests pass locally - thanks @hamaker

Issue #55, opened 2020-05-20
ec2_vol - add support for modifying existing volumes


briantist:
##### SUMMARY

This is a re-submission of https://github.com/ansible/ansible/issues/21588

A PR was opened to implement this but was ultimately closed without being merged seemingly due to disagreements over whether to require a complete port of the module to boto3:
https://github.com/ansible/ansible/pull/22978

The need for the feature, particularly for expanding existing volumes, is still very real though.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

ec2_vol

ansibullbot:
Files identified in the description:
* [`plugins/modules/ec2_vol.py`](https://github.com/ansible-collections/amazon.aws/blob/main/plugins/modules/ec2_vol.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @jillr @lwade @s-hertel @tremble @wimnat
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

roy-boy:
Any update on this one or we just waiting 2.10 to be released?

tremble:
@roy-boy As it stands the ec2_vol module is a boto (v2) based module and I believe the original boto python module doesn't support changing the size of volumes.  While it's possible to create a hybrid boto/boto3 module (as done in ansible/ansible#22978) this generally results in a much higher risk of bugs creeping into the module as boto and boto3 have very different approaches to parameters, return values and exceptions.

There's a port of this module to boto3 in progress (#53).  However, the bulk of contributors to both ansible.aws and community.aws are 'community' contributors and not directly paid to work on Ansible collections.  As such, turn around can be much slower than people may like.  Once #53 is merged reviving ansible/ansible#22978 should be relatively trivial.

roy-boy:
@tremble , thanks for your reply, yeah understand the release cycle of open source project. No worries I can use python boto3 now as the workaround for my project and keep an eye on the 2.10 release notes. Cheers!

markuman:
https://github.com/ansible-collections/amazon.aws/pull/215

markuman:
closed by https://github.com/ansible-collections/amazon.aws/pull/215

PR #215, opened 2020-12-06
ec2_vol support for gp3 volume

markuman:
##### SUMMARY

fixes also: https://github.com/ansible-collections/amazon.aws/issues/55

AWS introduced a new EBS volume type `gp3`.  
https://aws.amazon.com/about-aws/whats-new/2020/12/introducing-new-amazon-ebs-general-purpose-volumes-gp3/

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ec2_vol

##### ADDITIONAL INFORMATION

example playbook to transform all gp2 volumes with less than 3000 iops to gp3.

```yml
---
- hosts: localhost
  connection: local

  tasks:
    
    - name: ebs volume info
      amazon.aws.ec2_vol_info:
      register: EBS

    - name: ebs gp2 < 3000 iops
      debug:
        msg: "{{item.id}}"
      when: item.type == "gp2" and item.iops < 3000
      with_items: "{{ EBS.volumes }}"

    - name: change from gp2 with less 3000 iops to gp3
      ec2_vol:
        region: "{{ item.region }}"
        id: "{{ item.id }}"
        volume_type: gp3
        zone: "{{ item.zone }}"
      when: item.type == "gp2" and item.iops < 3000
      with_items: "{{ EBS.volumes }}"
```


markuman:
- [ ] expand integration test
  * modifiy existing volume
  * create new gp3 volume
- [ ] add new parameter `throughput` (_only available for gp3_)
- [ ] docs and changelog snippet

markuman:
So CI throws the same as befor I add my changes.

markuman:
I thought it would be easy to implement the new gp3 volume_type. Because I think it is a must have for many many users.  
Creating is not the problem.   
But modifiny....is a minefield!

Until now, `ec2_vol` does not consider that a volume can be modified.  
It becomes complicated when detaching a volume, without mention its current `volume_type`. Because the default `volume_type` in ansible is `standard`, detaching it like the integration test did, will modify the volume to ansible default value ('standard').

nowadays when you create an ebs volume, `gp2` is the default (console and boto3). So just create one, attach it somewhere and  detach it with ansible now. Here is where the problems are starting.

```yml
    - name: detach volume from the instance
      ec2_vol:
        id: vol-05062cb083be0c5cb
        instance: ""
      register: new_vol_attach_result_idem

    - name: detach volume from the instance (idempotent)
      ec2_vol:
        id: vol-05062cb083be0c5cb
        instance: ""
      register: new_vol_attach_result_idem
```

So basically I read out the original `volume_type`, which is `gp2`. And I compare it with the target volume type, which is requested by ansible. Because no volume type is defined, the volume type will be changed from gp2 to standard.  
The 2nd idempotent detach will fail, because the process is still in "modifying".  

So it is bad/dangerous to detach or attach a volume without mention the `volume_type`. It results in an unwanted change.

I see no solution to handle it when we keep the current behaviour.  
And I'm afraid we will drop previous behaviour of `ec2_vol` in any case.

When we do nothing, I need to change the integration test an add the missing `volume_type` to every task.  
Then the CI pass, but after releasing, it will result in unwanted changes for all users who are detach/attach volume without mention the volume_type.

One possible solution might be to control detach and attach with the `state` parameter. In this case, a none-present of the `volume_type` can be ignored.  
In this case, attaching and detaching will not work as before, because all use needs to control it with the `state` parameter.

Another opportunity is to drop the default `volume_type`.  
Attach and detach will work as before, but you cannot create a volume anymore without mention a volume_type. But I guess most users are defining one.

Any ideas @tremble @jillr

markuman:
the only solution I can think of is dropping the default value in line 604 https://github.com/ansible-collections/amazon.aws/pull/215/files#diff-001d1129c264b73f5eb92f50ebbab3c97ce8abeb9b99ae7e631d04e219fb0a4fL604

And in case the volume does not exist and no volume_type is defined, we use the default volume_type "standard".  
Line 751 https://github.com/ansible-collections/amazon.aws/pull/215/files#diff-001d1129c264b73f5eb92f50ebbab3c97ce8abeb9b99ae7e631d04e219fb0a4fR751

that keeps backwards compatibility in case of "creating" volumes (_the only thing the volume does atm_).

markuman:
As @flowerysong mentioned in irc. We can introduce a new parameter `modify_volume: no`.  
Maybe that's the cleanest solution.

markuman:
I've no idea why shippable CI is failing.  
My local tests runs successfull.  
any ideas @tremble @jillr @flowerysong  ?

```
ansible-test integration -v --python 3.7 ec2_vol --docker --allow-unsupported
...
testhost                   : ok=69   changed=22   unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   


ansible-test integration -v --python 2.7 ec2_vol --docker --allow-unsupported
...
testhost                   : ok=69   changed=21   unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

markuman:
> ModifyVolume operation: You are not authorized to perform this operation.

https://github.com/mattclay/aws-terminator/blob/c7f90bcb1905b14e65e26c04ed0365a0f150c37f/aws/policy/compute.yaml needs permissions to modify ebs volumes.

markuman:
Ready for review.

Sadly after `git mv tests/integration/targets/ec2_vol/tasks/main.yml tests/integration/targets/ec2_vol/tasks/tests.yml`, the tests.yml is detected as completely new. So here is the diff with the previous integration test.

```diff
--- /tmp/main.yml       2021-01-11 09:37:35.504606323 +0100
+++ tests/integration/targets/ec2_vol/tasks/tests.yml   2021-01-11 09:37:49.457939457 +0100
@@ -11,8 +11,6 @@
     - amazon.aws
 
   block:
-
-    # ==== Env setup ==========================================================
     - name: list available AZs
       aws_az_info:
       register: region_azs
@@ -314,6 +312,63 @@
         that:
           - not new_vol_attach_result_idem.changed
 
+    - name: must not change because of missing parameter modify_volume
+      ec2_vol:
+        id: "{{ new_vol_attach_result.volume_id }}"
+        zone: "{{ availability_zone }}"
+        volume_type: gp3
+      register: changed_gp3_volume
+
+    - name: volume must not changed
+      assert:
+        that:
+          - not changed_gp3_volume.changed
+
+    - name: change existing volume to gp3
+      ec2_vol:
+        id: "{{ new_vol_attach_result.volume_id }}"
+        zone: "{{ availability_zone }}"
+        volume_type: gp3
+        modify_volume: yes
+      register: changed_gp3_volume
+
+    - name: check that volume_type has changed
+      assert:
+        that:
+          - changed_gp3_volume.volume_type == 'gp3'
+          - changed_gp3_volume.changed
+
+    - name: volume must be from type gp3 (idempotent)
+      ec2_vol:
+        id: "{{ new_vol_attach_result.volume_id }}"
+        zone: "{{ availability_zone }}"
+        volume_type: gp3
+        modify_volume: yes
+      register: changed_gp3_volume
+      retries: 3
+      delay: 3
+      until: not changed_gp3_volume.failed
+      # retry because ebs change is to slow
+
+    - name: must not changed (idempotent)
+      assert:
+        that:
+          - changed_gp3_volume.volume_type == 'gp3'
+          - not changed_gp3_volume.changed
+
+    - name: re-read volume information to validate new volume_type
+      ec2_vol_info:
+        filters:
+          volume-id: "{{ changed_gp3_volume.volume_id }}"
+      register: verify_gp3_change
+
+    - name: volume type must be gp3
+      assert:
+        that:
+          - v.type == 'gp3'
+      vars:
+        v: "{{ verify_gp3_change.volumes[0] }}"
+
     - name: delete volume
       ec2_vol:
         id: "{{ volume2.volume_id }}"
@@ -399,19 +454,44 @@
           v: "{{ volume_info.volumes[0] }}"
       when: ansible_version.full is version('2.7', '>=')
 
+    - name: test create a new gp3 volume
+      ec2_vol:
+        volume_size: 1
+        volume_type: gp3
+        zone: "{{ availability_zone }}"
+        throughput: 130
+        tags:
+          ResourcePrefix: "{{ resource_prefix }}"
+      register: gp3_volume
 
-    # ==== Cleanup ============================================================
+    - name: check that volume_type is gp3
+      assert:
+        that:
+          - gp3_volume.volume_type == 'gp3'
+          - gp3_volume.changed
+          - gp3_volume.volume.throughput == 130
 
-  always:
-    - name: Describe the volume before we delete it
-      ec2_vol_info:
-        id: "{{ item.volume_id }}"
-      ignore_errors: yes
-      register: pre_delete
+    - name: increase throughput
+      ec2_vol:
+        volume_size: 1
+        volume_type: gp3
+        zone: "{{ availability_zone }}"
+        throughput: 131
+        tags:
+          ResourcePrefix: "{{ resource_prefix }}"
+      register: gp3_volume
+
+    - name: check that throughput has changed
+      assert:
+        that:
+          - gp3_volume.volume_type == 'gp3'
+          - gp3_volume.changed
+          - gp3_volume.volume.throughput == 131
 
-    - debug:
-        var: pre_delete
 
+    # ==== Cleanup ============================================================
+
+  always:
     - name: Describe the instance before we delete it
       ec2_instance_info:
         instance_ids:
@@ -441,6 +521,7 @@
         - "{{ new_vol_attach_result }}"
         - "{{ attach_new_vol_from_snapshot_result }}"
         - "{{ dot_volume }}"
+        - "{{ gp3_volume }}"
 
     - name: delete snapshot
       ec2_snapshot:
```

cc @jillr @tremble @flowerysong @goneri @gravesm

todori438:
Hi @markuman,
Thanks for the great initiative, it will definitely make our lives easier!
I just tried to run an ansible playbook with gp3 as volume_type, but it ended with an error: {"changed": false, "msg": "value of volume_type must be one of: standard, gp2, io1, st1, sc1, got: gp3"}

    - name: must not change because of missing parameter modify_volume
      ec2_vol:
        id:  vol-xxx
        volume_type: gp3

Any idea what needs to be done from my side to use gp3 as value?

ansible --version
ansible 2.9.16
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /bin/ansible
  python version = 2.7.5 (default, Apr  2 2020, 13:16:51) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]

markuman:
@todori438 

> ` {"changed": false, "msg": "value of volume_type must be one of: standard, gp2, io1, st1, sc1, got: gp3"}`

That looks like that the origin `ec2_vol` module what ansible 2.9.16 shipps is taken and not the one from this PR.  
You can verify this by running your playbook with `-vvv`.

Where do you've placed the playbook? When you want to "hot-load" it, you can place it in a folder named `library/` at the same level of your playbook.  
Another opportunity is to checkout the entire PR and run `ansible-galaxy collection build` and `ansible-galaxy collection install <filename>.tar.gz`.

todori438:
Hi @markuman,

Thanks for the help.
I couldn't make it working :( 
What i did:
git clone https://github.com/markuman/amazon.aws.git
Cloning into 'amazon.aws'...
remote: Enumerating objects: 15, done.
remote: Counting objects: 100% (15/15), done.
remote: Compressing objects: 100% (11/11), done.
remote: Total 2729 (delta 5), reused 12 (delta 4), pack-reused 2714
Receiving objects: 100% (2729/2729), 1.01 MiB | 0 bytes/s, done.
Resolving deltas: 100% (1795/1795), done.

 amazon.aws]# ansible-galaxy collection build
[WARNING]: Found unknown keys in collection galaxy.yml at
'/amazon.aws/galaxy.yml': build_ignore
Created collection for amazon.aws at /amazon.aws/amazon-aws-1.2.1.tar.gz

ansible-galaxy collection install amazon-aws-                            1.2.1.tar.gz
Process install dependency map
Starting collection install process
Installing 'ansible.netcommon:1.4.1' to '/root/.ansible/collections/ansible_collections/ansible/netcommon'
Installing 'amazon.aws:1.2.1' to '/root/.ansible/collections/ansible_collections/amazon/aws'

Then, ran the same playbook and got same error:
TASK [must not change because of missing parameter modify_volume] ******************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "value of volume_type must be one of: standard, gp2, io1, st1, sc1, got: gp3"}

Currently the playbook is located under the same directory where i did the pull of your git repo
/amazon.aws

Could you please advise what i'm doing wrong ?
Sorry for these stupid questions but i cannot make it working :(

gravesm (approved):
@markuman Thanks for all your work on this!

goneri (approved):

plugins/modules/ec2_vol.py
```diff
@@ -91,11 +92,25 @@
       - tag:value pairs to add to the volume after creation.
     default: {}
     type: dict
+  modify_volume:
+    description:
+      - Ony when set to C(true), the volume will be modified.
```
goneri, line 97: I would just say: "The volume won't be modify unless this key is C(true)".

jillr (approved):
Thanks for all your work on this @markuman

jillr:
The only change since goneri's review is the requested text, going to go ahead and merge

Issue #235, opened 2021-01-14
AWS inverse filter in Ansible 2.10.5 amazon aws ec2.py causing bad return and empty array during stop task. 


IamNobleSix:
##### SUMMARY
In PR https://github.com/ansible-collections/amazon.aws/pull/197/files there was an addition to ec2.py which added an inverse lookup to remove terminated and shutting-down instances from the instance-state-name filter. To the best of my knowledge, aws ec2 filters do not support inverse filters. This is causing a bad return from aws ec2, which removes all hosts from the actionable command. 

Our ec2 instance stop playbook broke when moving from ansible 2.10.4 to 2.10.5, and the file diff showed the filter line as the only impacting change in ec2.py. By removing this inverse filter the playbooks worked again. This is the diff between the working ec2.py and NOT working ec2.py

```
Desktop:$ diff working_ec2_py.txt broken_ec2_py.txt
1418a1419,1420
>     filters['instance-state-name'] = ["!terminated", "!shutting-down"]
>
1739d1740
<
Desktop:$
```

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
amazon aws ec2 module

##### ANSIBLE VERSION

```
~:$ ansible --version
ansible 2.10.4
  config file = None
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.10.5/libexec/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.1 (default, Jan  8 2021, 17:17:17) [Clang 12.0.0 (clang-1200.0.32.28)]
```

##### CONFIGURATION

```
~:$ ansible-config dump --only-changed
ANSIBLE_NOCOWS(env: ANSIBLE_NOCOWS) = True
```

##### OS / ENVIRONMENT
macOS Catalina 10.15.7

##### STEPS TO REPRODUCE

Use 2.10.4 amazon aws ec2.py module and attempt to stop a running instance. Then use 2.10.5 amazon aws ec2.py module and try again. We are simply providing an instance_id via our inventory for the stop. 

```yaml
  - name: Check if Instance is already Stopped
    delegate_to: 127.0.0.1
    ec2_instance_info:
      validate_certs: no
      aws_access_key: "{{ aws_access_key | default(omit) }}"
      aws_secret_key: "{{ aws_secret_key | default(omit) }}"
[...]

PR #237, opened 2021-01-14
remove instance-state-name inverse filters in favor of standard value…

IamNobleSix:
remove instance-state-name inverse filters in favor of standard value filters on other 4 accepted options to fix bad aws return

##### SUMMARY

This fixes: https://github.com/ansible-collections/amazon.aws/issues/235

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

amazon aws ec2.py

##### ADDITIONAL INFORMATION
Instead of using inverse filters (which to my knowledge are not allowed by any or all version of aws cli and can cause issues as noted in https://github.com/ansible-collections/amazon.aws/issues/235) use the other 4 allowed values as normal outlined in https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instances.html for instance-state-name

### DIFF
--- a/changelogs/fragments/237_replace_inverse_ec2_aws_filter.yaml
+++ b/changelogs/fragments/237_replace_inverse_ec2_aws_filter.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+- aws_ec2 module - Replace inverse aws instance-state-name filters !terminated, !shutting-down in favor of postive filters pending, running, stopping, stopped. Issue 235. (https://github.com/ansible-collections/amazon.aws/pull/237)
--- a/plugins/modules/ec2.py
+++ b/plugins/modules/ec2.py
@@ -1416,7 +1416,7 @@ def startstop_instances(module, ec2, instance_ids, state, instance_tags):
         for key, value in instance_tags.items():
             filters["tag:" + key] = value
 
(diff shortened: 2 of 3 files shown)


gravesm (changes requested):
Thank you for the PR, @IamNobleSix. This change looks good. We need a changelog fragment before we can merge: https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment.

It would also be a good idea to add a test for this, which would have caught this regression before it went out. In this case, I think just adding a task to stop the second instance here: https://github.com/ansible-collections/amazon.aws/blob/main/tests/integration/targets/ec2/tasks/main.yml#L138 and then assert that it is in fact stopped should be sufficient.

IamNobleSix:
Hi @gravesm that sounds good to me. I will get working on that today. Thanks.

IamNobleSix:
Stop test created using similar logic to already established terminate test on line 103.

gravesm (approved):
Thanks for your work on this, @IamNobleSix!

goneri (approved):

jillr (approved):

goneri:
Thank you @IamNobleSix for your contribution.

PR #278, opened 2021-03-04
drop the a.netcommon dependency

goneri:
Remove the dependency on the `ansible.netcommon` dependency.

Closes: #274
Closes: #273

PR #280, opened 2021-03-04
address pylint errors

goneri:
This commit addresses the three following pylint errors:

- plugins/module_utils/cloud.py:145:27: isinstance-second-argument-not-valid-type: Second argument of isinstance is not a type
- plugins/module_utils/cloudfront_facts.py:166:0: unnecessary-comprehension: Unnecessary use of a comprehension
- plugins/module_utils/core.py:192:4: property-with-parameters: Cannot have defined parameters for properties


goneri:
Merged in https://github.com/ansible-collections/amazon.aws/pull/278

PR #274, opened 2021-03-03
galaxy: Fix ansible.common collection version

Akasurde:
##### SUMMARY

ansible.common removed ipaddress which is required by ec2_group module.
Fix ansible.common version <2.0.0 to handle this dependency.

Fixes: #273

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
changelogs/fragments/fix_netcommon.yml
galaxy.yml

PR #281, opened 2021-03-04
drop the a.netcommon dependency

goneri:
Remove the dependency on the `ansible.netcommon` dependency.

Closes: #274


goneri:
It was a dup with https://github.com/ansible-collections/amazon.aws/pull/278

PR #282, opened 2021-03-04
Vendor ipaddress.py module util

### LINKED ISSUE #273: ipaddress module is removed in ansible.netcommon >=2.0.0
There was a major version bump yesterday on ansible.netcommon and now this is broken.

https://github.com/ansible-collections/ansible.netcommon/commit/8cda0e6906a86790ab423b11adafa425e870926a

https://github.com/ansible-collections/amazon.aws/blob/76aeec280994454fc2e99fcf3996e5a5ab1e8184/plugins/modules/ec2_group.py#L410 and perhaps other locations.

jillr:
##### SUMMARY
It had been previously vendored in ansible.netcommon but dropped in 2.0
This is needed for ec2_group until we can switch to python3-only, for ipv6 support.

Also drop ansible.netcommon dependency.

We can look at other options in the medium to long term, but in the shortest term this will address the issue with probably the fewest issues (needing to devise our own checking, being difficult for folks who need netcommon 2.0+ for other reasons, etc).

A point release will be prepared once this lands.

Fixes #273

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
module_utils
ec2_group

### DIFF
--- a/changelogs/fragments/273-vendor-ipaddress-utility.yml
+++ b/changelogs/fragments/273-vendor-ipaddress-utility.yml
@@ -1,0 +1,2 @@
+minor_changes:
+- module_utils - the ipaddress module utility has been vendored into this collection.  This eliminates the collection dependency on ansible.netcommon (which had removed the library in its 2.0 release).  The ipaddress library is provided for internal use in this collection only. (https://github.com/ansible-collections/amazon.aws/issues/273)-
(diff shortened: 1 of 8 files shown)


nitzmahone (approved):
LGTM- best possible solution I can think of for the moment

jillr:
there are lingering pylint issues - I'll deal with those separately as we need to get out of this chicken and egg CI failures situation and I'd like to do it in a way that is easy to revert if any one piece needs to be reverted at some point.

Issue #233, opened 2021-01-13
Add s3 object tagging support


QCU87Z:
##### SUMMARY

Currently I can not see away to add tags to S3 objects. The project I am on requires this feature

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

amazon.aws.aws_s3

##### ADDITIONAL INFORMATION

There should be another paramater for the amazon.aws.aws_s3 task that allows to you pass a dict of tags and their values

```yaml
- name: upload s3 boject
  amazon.aws.aws_s3
    bucket: input-bucket
    object: newObject
    src: files/object
    tags:
      - tag1: value1
      - tag2: value 2
```
Or something like that

ansibullbot:
Files identified in the description:
None

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

goneri:
@QCU87Z can you give a try to https://github.com/ansible-collections/amazon.aws/pull/335? The PR is mostly ready.

PR #335, opened 2021-04-22
aws_s3: add S3 object tagging feature 

alinabuzachis:
Depends-On: https://github.com/ansible-collections/amazon.aws/pull/372

##### SUMMARY

aws_s3: Add tagging feature
Fixes: https://github.com/ansible-collections/amazon.aws/issues/233

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

aws_s3

Requires: https://github.com/mattclay/aws-terminator/pull/146

### DIFF
--- a/plugins/modules/aws_s3.py
+++ b/plugins/modules/aws_s3.py
@@ -530,8 +530,6 @@ def option_in_extra_args(option):
 
 
 def upload_s3file(module, s3, bucket, obj, expiry, metadata, encrypt, headers, src=None, content=None):
-    result = {}
-
     if module.check_mode:
         module.exit_json(msg="PUT operation skipped - running in check mode", changed=True)
     try:
--- a/plugins/modules/aws_s3.py
+++ b/plugins/modules/aws_s3.py
@@ -155,6 +155,22 @@ options:
     description:
       - KMS key id to use when encrypting objects using I(encrypting=aws:kms). Ignored if I(encryption) is not C(aws:kms).
     type: str
+<<<<<<< HEAD
+=======
+  tags:
+    description:
+      - Tags dict to apply to the S3 object.
+    type: dict
+    version_added: 1.5.0
+  purge_tags:
+    description:
+      - Whether or not to remove tags assigned to the DB cluster if not specified in the playbook.
+      - To remove all tags set I(tags) to an empty dictionary in conjunction with this.
+    type: bool
+    default: True
+    version_added: 1.5.0
+requirements: [ "boto3", "botocore" ]
+>>>>>>> 	* Docs and logic fix
 author:
     - "Lester Wade (@lwade)"
     - "Sloane Hertel (@s-hertel)"
@@ -302,6 +318,7 @@ import os
 import io
 from ssl import SSLError
 import base64
+import time
 
 try:
     import botocore
@@ -318,6 +335,8 @@ from ..module_utils.core import is_boto3_error_message
 from ..module_utils.ec2 import AWSRetry
 from ..module_utils.ec2 import boto3_conn
 from ..module_utils.ec2 import get_aws_connection_info
+from ..module_utils.ec2 import ansible_dict_to_boto3_tag_list
+from ..module_utils.ec2 import boto3_tag_list_to_ansible_dict
 from ..module_utils.s3 import HAS_MD5
 from ..module_utils.s3 import calculate_etag
 from ..module_utils.s3 import calculate_etag_content
@@ -511,6 +530,9 @@ def option_in_extra_args(option):
 
 
 def upload_s3file(module, s3, bucket, obj, expiry, metadata, encrypt, headers, src=None, content=None):
+    tags = module.params.get("tags")
+    purge_tags = module.params.get("purge_tags")
+
     if module.check_mode:
         module.exit_json(msg="PUT operation skipped - running in check mode", changed=True)
     try:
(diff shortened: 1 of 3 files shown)


jillr (commented):
@alinabuzachis  looks like there's some logic missing, I wasn't able to test the backoff problem (do you have local changes that need to be pushed?). Here's a few I found.

plugins/modules/aws_s3.py
```diff
@@ -669,6 +703,42 @@ def get_s3_connection(module, aws_connect_kwargs, location, rgw, s3_url, sig_4=F
...
 
+def get_current_object_tags_dict(s3, bucket, obj):
+    try:
+        current_tags = s3.get_object_tagging(Bucket=bucket, Key=obj).get('TagSet')
+    except is_boto3_error_code('NoSuchTagSet'):
+        return {}
+    except is_boto3_error_code('NoSuchTagSetError'):  # pylint: disable=duplicate-except
+        return {}
+
+    return boto3_tag_list_to_ansible_dict(current_tags)
```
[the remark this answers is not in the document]
alinabuzachis: @jillr I was quite sure I pushed the latest changes. I'm sorry for that.

goneri:
recheck

goneri:
recheck

goneri:
recheck

jillr:
recheck

plugins/modules/aws_s3.py
```diff
@@ -156,6 +156,18 @@
     description:
       - KMS key id to use when encrypting objects using I(encrypting=aws:kms). Ignored if I(encryption) is not C(aws:kms).
     type: str
+  tags:
+    description:
+      - Tags dict to apply to the S3 object.
+    type: dict
+    version_added: 1.5.0
```
jillr, line 163:
```suggestion
    version_added: 2.0.0
```

plugins/modules/aws_s3.py
```diff
@@ -156,6 +156,18 @@
...
+  purge_tags:
+    description:
+      - Whether or not to remove tags assigned to the DB cluster if not specified in the playbook.
+      - To remove all tags set I(tags) to an empty dictionary in conjunction with this.
+    type: bool
+    default: True
+    version_added: 1.5.0
```
jillr, line 170:
```suggestion
    version_added: 2.0.0
```

plugins/modules/aws_s3.py
```diff
@@ -156,6 +156,18 @@
...
       - KMS key id to use when encrypting objects using I(encrypting=aws:kms). Ignored if I(encryption) is not C(aws:kms).
     type: str
+  tags:
+    description:
+      - Tags dict to apply to the S3 object.
+    type: dict
+    version_added: 1.5.0
+  purge_tags:
+    description:
+      - Whether or not to remove tags assigned to the DB cluster if not specified in the playbook.
```
jillr, line 166:
```suggestion
      - Whether or not to remove tags assigned to the S3 object if not specified in the playbook.
```

jillr (commented):
This looks good, for `put` modes I think we should just call `ensure_tags()` after PUTing the object, rather than do a bunch of data transformation to provide the tags as ` Tagging='key1=value1&key2=value2',`

plugins/modules/aws_s3.py
alinabuzachis: > This looks good, for `put` modes I think we should just call `ensure_tags()` after PUTing the object, rather than do a bunch of data transformation to provide the tags as ` Tagging='key1=value1&key2=value2',`

Thank you for reviewing @jillr. Ok, I was also thinking to do it in this way.

alinabuzachis:
> This looks good, for `put` modes I think we should just call `ensure_tags()` after PUTing the object, rather than do a bunch of data transformation to provide the tags as ` Tagging='key1=value1&key2=value2',`

Thank you for reviewing @jillr. Ok, I was also thinking to do it in this way.

jillr (approved):
LGTM, @goneri did you still want to review?

jillr:
Ok to gate once the merge conflict is resolved

goneri:
recheck

goneri:
recheck

jillr:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

alinabuzachis:
recheck

plugins/modules/aws_s3.py
```diff
@@ -700,7 +814,9 @@ def main():
         content=dict(),
         content_base64=dict(),
         ignore_nonexistent_bucket=dict(default=False, type='bool'),
-        encryption_kms_key_id=dict()
+        encryption_kms_key_id=dict(),
+        tags=dict(type='dict'),
+        purge_tags=dict(type='bool', default=True),
```
bedge, line 819: Why was this deemed a good idea?
Now, everything that touches any asset that uses tags MUST now add a `purge_tags: False` just to leave things as they are??

Use case:
```
- community.aws.ec2_asg:
    name: "{{ item.auto_scaling_group_name }}"
    min_size: "{{ count }}"
    max_size: "{{ count }}"
    desired_capacity: "{{ count }}"
    region: eu-central-1
```

Used to allow modifying an ASG min/max/desired size.

Now, does the above and **strips all tags**, how is that the "principle of least surprising behavior"?

This requires adding the `purge_tags: False` **everywhere** where tags could be impacted.

```
- community.aws.ec2_asg:
    name: "{{ item.auto_scaling_group_name }}"
    min_size: "{{ count }}"
    max_size: "{{ count }}"
    desired_capacity: "{{ count }}"
    purge_tags: False
    region: eu-central-1
```

This is the ONLY case where one has to specify a non-default option to "leave things alone".

PR #340, opened 2021-04-26
test/s3: keep the bucket name length under control (<63 characters)

goneri:
The bucket_name variable length is 63 or below. If we add a prefix we
can generate an invalid bucket name.

See: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html

### DIFF
--- a/tests/integration/targets/aws_s3/tasks/main.yml
+++ b/tests/integration/targets/aws_s3/tasks/main.yml
@@ -444,7 +444,7 @@
 
     - name: test create a bucket with a dot in the name
       aws_s3:
-        bucket: "{{ bucket_name + '.bucket' }}"
+        bucket: "{{ bucket_name | hash('md5') + '.bucket' }}"
         mode: create
       register: result
 
@@ -454,7 +454,7 @@
 
     - name: test delete a bucket with a dot in the name
       aws_s3:
-        bucket: "{{ bucket_name + '.bucket' }}"
+        bucket: "{{ bucket_name | hash('md5') + '.bucket' }}"
         mode: delete
       register: result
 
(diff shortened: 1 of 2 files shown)


jillr:
While it's probably safe to drop these tests, I'd generally prefer not to drop tests if possible.  We'll probably also see resource name length issues in other test suites, are we not able to control the names we get from zuul?

goneri:
> While it's probably safe to drop these tests, I'd generally prefer not to drop tests if possible. We'll probably also see resource name length issues in other test suites, are we not able to control the names we get from zuul?

The prefix is generated by `ansible-test`. Follow up to this two PR what was a de facto convention is now enforced. The prefix is 63 character long.
- https://github.com/ansible/ansible/pull/73826
- https://github.com/ansible/ansible/pull/73844

So we're good as soon as we don't had anything to the prefix.

Regarding this specific case, I was under the impression the test is a bit pointless. I'm since updated my patch to continue to run it. I now build the bucket name with the MD5 of the prefix.

jillr:
The test might be pointless, it seems likely.  When it was written the module was boto2 so it's entirely possible it was addressing some case that's no longer a concern. But I couldn't find any more info about why we wrote that test, like if there's a history of S3/the SDK doing weird things with those bucket names.  So I'm hesitant to remove the tests just to be conservative.

goneri:
> The test might be pointless, it seems likely. When it was written the module was boto2 so it's entirely possible it was addressing some case that's no longer a concern. But I couldn't find any more info about why we wrote that test, like if there's a history of S3/the SDK doing weird things with those bucket names. So I'm hesitant to remove the tests just to be conservative.

Sorry for the harsh tone of my initial comment :-). I started a PR to validate the name of the  bucket: https://github.com/ansible-collections/amazon.aws/pull/341

jillr:
No worries, no hard tone was heard!  :)

PR #345, opened 2021-04-27
s3_bucket/test: use one bucket per scenario

goneri:
To speed up the execution, we run all the tests in parallel. This commit
ensures each playbook uses a different bucket, this to avoid any
unexpected side-effect.

### DIFF
--- a/tests/integration/targets/s3_bucket/roles/s3_bucket/tasks/complex.yml
+++ b/tests/integration/targets/s3_bucket/roles/s3_bucket/tasks/complex.yml
@@ -1,8 +1,10 @@
 ---
 - block:
+    - set_fact:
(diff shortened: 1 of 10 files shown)


jillr (approved):
Thanks @goneri!

tremble (commented):
There's a much less invasive option:

in tests/integration/targets/s3_bucket/roles/s3_bucket/defaults.yml

```
bucket_name: '{{ resource_prefix | hash("md5") }}{{ suffix_override | default("-" + inventory_hostname) }}'
```

For dotted we can set update the inventory:
tests/integration/targets/s3_bucket/inventory
```
dotted suffix_override=".dotted"
```

tremble (approved):
since the change works let's get this merged to unblock #311

Issue #308, opened 2020-12-16
ec2 module documentation: no "Return values" section


selivan:
##### SUMMARY
ec2 module documentation does not have "Return values" section: https://docs.ansible.com/ansible/latest/collections/amazon/aws/ec2_module.html

So if you want to use this module results in later actions, you have to run it once and check manually what does it print. Not very convenient.

##### ISSUE TYPE
- Documentation Report

##### COMPONENT NAME

ec2

##### ANSIBLE VERSION

```paste below
ansible 2.10.3
```

Issue moves from here: https://github.com/ansible/ansible/issues/41220 because now this is separate repository

goneri:
Hi @selivan,

Thank you for the bug report, would you like to submit a PR to improve the documentation?

ansibullbot:
Files identified in the description:
* [`plugins/modules/ec2.py`](https://github.com/['ansible-collections/amazon.aws', 'ansible-collections/community.aws', 'ansible-collections/community.vmware']/blob/main/plugins/modules/ec2.py)

If these files are inaccurate, please update the `component name` section of the description or use the `!component` bot command.

[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

ansibullbot:
cc @jillr @lwade @s-hertel @tgerla @tremble @wimnat
[click here for bot help](https://github.com/ansible/ansibullbot/blob/master/ISSUE_HELP.md)

PR #364, opened 2021-05-10
ec2: Update documentation

alinabuzachis:
##### SUMMARY

Document returned values.
Fixes: https://github.com/ansible-collections/amazon.aws/issues/308

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

ec2

### DIFF
--- a/plugins/modules/ec2.py
+++ b/plugins/modules/ec2.py
@@ -610,10 +610,10 @@ instances:
                 }
             }
         dns_name:
-            description: If the EC2 instance has changed.
+            description: The public DNS name assigned to the instance.
             type: str
             returned: always
-            sample: ""
+            sample: "ec2-203-0-113-1.z-2.compute-1.amazonaws.com"
         ebs_optimized:
             description: Indicates whether the instance is optimized for Amazon EBS I/O.
             type: bool
@@ -684,12 +684,12 @@ instances:
             description: The public DNS name assigned to the instance.
             type: str
             returned: always
-            sample: ""
+            sample: "ec2-203-0-113-1.z-2.compute-1.amazonaws.com"
         public_ip:
             description: The public IPv4 address, or the Carrier IP address assigned to the instance, if applicable.
             type: str
             returned: always
-            sample: ""
+            sample: "203.0.113.1"
         ramdisk:
             description: The RAM disk associated with this instance, if applicable.
             type: str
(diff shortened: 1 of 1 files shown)


plugins/modules/ec2.py
```diff
@@ -577,6 +577,297 @@
...
+            returned: always
+            sample: {
+                "/dev/xvda": {
+                    "delete_on_termination": true,
+                    "status": "attached",
+                    "volume_id": "vol-06d364586f5550b62"
+                }
+            }
+        dns_name:
+            description: If the EC2 instance has changed.
```
jillr, line 613:
```suggestion
            description: The public DNS name assigned to the instance.
```
I'm reasonably certain it's the public address at least, http://boto.cloudhackers.com/en/latest/ref/ec2.html#boto.ec2.instance.Instance doesn't document dns_name.

plugins/modules/ec2.py
```diff
@@ -577,6 +577,297 @@
...
+            type: str
+            returned: always
+            sample: ""
```
jillr, line 616:
```suggestion
            sample: "ec2-203-0-113-1.z-2.compute-1.amazonaws.com"
```
203.0.113.0/24 is a reserved block that can safely be used in documentation.  The same can be used for public_dns_name below.

plugins/modules/ec2.py
```diff
@@ -577,6 +577,297 @@
...
+            sample: {
+                "capacity_reservation_preference": "open"
+            }
+        client_token:
+            description: The idempotency token you provided when you launched the instance, if applicable.
+            type: str
+            returned: always
+            sample: ""
+        cpu_options:
+            description: The idempotency token you provided when you launched the instance, if applicable.
```
jillr, line 753:
```suggestion
            description: The CPU options for the instance.
```

jillr (commented):
Thanks @alinabuzachis, just a couple small comments!

plugins/modules/ec2.py
alinabuzachis: Yes, you're right. In the copy/paste process I forget to rectify. Thank you.

plugins/modules/ec2.py
alinabuzachis: @jillr Thank you. Just added it (even for public_dns_name).

plugins/modules/ec2.py
alinabuzachis: @jillr Thanks.

jillr (approved):

Issue #120, opened 2020-07-30
Type error on no_device parameter of ec2_ami module


JulienSchneider:
##### SUMMARY

I am using the `no_device: yes` parameter of the ec2_ami module as described in the documentation. Ansible complain about that parameter type:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: Invalid type for parameter BlockDeviceMappings[1].NoDevice, value: True, type: <class 'bool'>, valid types: <class 'str'>
fatal: [localhost]: FAILED! => {"boto3_version": "1.14.29", "botocore_version": "1.17.29", "changed": false, "msg": "Error registering image: Parameter validation failed:\nInvalid type for parameter BlockDeviceMappings[1].NoDevice, value: True, type: <class 'bool'>, valid types: <class 'str'>"}
```

I am able to use that parameter when using  `no_device: ""`

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

ec2_ami

##### ANSIBLE VERSION

```
ansible 2.9.10
  config file = /home/xxx/ansible.cfg
  configured module search path = ['/home/xxx/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/xxx/.pyenv/versions/3.7.2/lib/python3.7/site-packages/ansible
  executable location = /home/xxx/.pyenv/versions/3.7.2/bin/ansible
  python version = 3.7.2 (default, Jul  3 2020, 10:51:01) [GCC 9.3.0]

```

##### CONFIGURATION

```paste below
$ ansible-config dump --only-changed
DEFAULT_HOST_LIST(/home/xxx/ansible.cfg) = ['/home/xxx/hosts']
INTERPRETER_PYTHON(/home/xxx/ansible.cfg) = auto
```

##### OS / ENVIRONMENT

Ubuntu 20.04 LTS

##### STEPS TO REPRODUCE

Run the test playbook below

```yaml
---

- hosts: localhost
  connection: local
  tasks:
    - name: Register AMI
      ec2_ami:
        instance_id: "i-xxxxxxxx"
        name: "graphite_ami_{{ ansible_date_time.iso8601_basic_short }}"
        device_mapping:
          - device_name: /dev/sda1
            volume_size: 10
            delete_on_termination: true
            volume_type: gp2
          - device_name: /dev/sdf
            no_device: yes
        tags:
          Name: Graphite

```

##### EXPECTED RESULTS

[...]

PR #386, opened 2021-06-14
Add fix to change the no_device parameter value to string 

srirachanaachyuthuni:
##### SUMMARY

Changes the value of no_device parameter from Bool to String for Boto3

Fixes: #120 
##### ISSUE TYPE

- Bugfix Pull Request - Type error on no_device parameter of ec2_ami module #120

##### COMPONENT NAME

ec2_ami
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/ec2_ami.py
+++ b/plugins/modules/ec2_ami.py
@@ -474,6 +474,8 @@ def create_image(module, connection):
                 device = rename_item_if_exists(device, 'iops', 'Iops', 'Ebs')
                 device = rename_item_if_exists(device, 'encrypted', 'Encrypted', 'Ebs')
 
+                # The NoDevice parameter in Boto3 is a string. Empty string omits the device from block device mapping
+                # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.create_image
                 if 'NoDevice' in device:
                     if device['NoDevice'] is True:
                         device['NoDevice'] = ""
--- a/plugins/modules/ec2_ami.py
+++ b/plugins/modules/ec2_ami.py
@@ -457,7 +457,6 @@ def create_image(module, connection):
         }
 
         block_device_mapping = None
-
         # Remove empty values injected by using options
         if device_mapping:
             block_device_mapping = []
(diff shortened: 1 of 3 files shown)


plugins/modules/ec2_ami.py
```diff
@@ -474,6 +473,12 @@ def create_image(module, connection):
                 device = rename_item_if_exists(device, 'volume_size', 'VolumeSize', 'Ebs', attribute_type=int)
                 device = rename_item_if_exists(device, 'iops', 'Iops', 'Ebs')
                 device = rename_item_if_exists(device, 'encrypted', 'Encrypted', 'Ebs')
+
+                if 'NoDevice' in device:
+                    if device['NoDevice'] is True:
```
jillr, line 478: Could you add a brief comment here explaining why this is being done?  Maybe even just a link to the boto or API reference, since this option is a little odd.

jillr (commented):
Tiny nit, otherwise this is looking pretty good - thanks @srirachanaachyuthuni!

jillr:
@abikouo Could you give this PR a review when you have a chance please?

tests/integration/targets/ec2_ami/tasks/main.yml
```diff
@@ -112,6 +112,53 @@
...
+      register: result_no_device_false
+
+    - name: set image id fact for deletion later
+      set_fact:
+        ec2_ami_no_device_false_image_id: "{{ result_no_device_false.image_id }}"
+
+    - name: assert that image with no_device option yes and no_device option no has been created
+      assert:
+        that:
+          - "result_no_device_true.changed"
```
abikouo, line 157: it is better to check the task output just after the task itself instead of having an assert block testing several tasks, in that way you validate that each use case was successfully validated before running with the next one

abikouo (changes requested):

tests/integration/targets/ec2_ami/tasks/main.yml
abikouo, line 157: testing that there was a changed is good for what was expected in the issue, is there a way to test that the change is as expected ?? using ``ec2_ami_info``

srirachanaachyuthuni: I checked that the result is as expected by  checking for the device mapping in the output recieved after creating ami rather than using the ec2_ami_info module to make it simpler and have fewer lines of code.

tests/integration/targets/ec2_ami/tasks/main.yml
```diff
@@ -112,6 +112,59 @@
...
+      register: result_no_device_true
+
+    - name: set image id fact for deletion later
+      set_fact:
+        ec2_ami_no_device_true_image_id: "{{ result_no_device_true.image_id }}"
+
+    - name: assert that image with no_device option yes has been created
+      assert:
+        that:
+          - "result_no_device_true.changed"
```
abikouo, line 138: quote are not necessary here (same for line 163)

abikouo (approved):
LGTM
Great job!!

jillr (approved):

PR #378, opened 2021-05-31
aws_ssm add on_missing and on_denied options

abikouo:
##### SUMMARY

closes #126

##### ISSUE TYPE

- Feature Pull Request
##### COMPONENT NAME

aws_ssm
##### ADDITIONAL INFORMATION

```paste below
- set_fact:
     param: "{{ lookup('amazon.aws.aws_ssm', 'variable', on_missing='error', on_denied='warn')}}"
```

PR #388, opened 2021-06-16
Fix versioned parameters

dginther:
##### SUMMARY

Fixes an issue where a versioned parameter in the list causes a KeyError.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ssm.py

##### ADDITIONAL INFORMATION

Problem:
If you request a versioned parameter using the lookup function, you get an unhandled exception. This is because although we are checking for just the name of the parameter (by using .split), when we add it to the list we are returning, we use the name as-is for the index lookup.

```
session = boto3.Session()
ssm = session.client('ssm')
ssm_dict["Names"] = ['/path/to/parameter:1']
response = ssm.get_parameters(**ssm_dict)
params = boto3_tag_list_to_ansible_dict(response['Parameters'], tag_name_key_name="Name",
                                                                    tag_value_key_name="Value")
for i in ['/path/to/parameter:1']:
  if i.split(':', 1)[0] in params:
    print(params[i])

  File "<stdin>", line 3, in <module>
KeyError: '/path/to/parameter:1'
```

To solve this problem, we need to set `i` to the value without the version identifier. This works regardless of whether the parameter is versioned or not.

Solution:
```
session = boto3.Session()
ssm = session.client('ssm')
ssm_dict["Names"] = ['/path/to/parameter:1']
response = ssm.get_parameters(**ssm_dict)
params = boto3_tag_list_to_ansible_dict(response['Parameters'], tag_name_key_name="Name",
                                                                    tag_value_key_name="Value")
for i in ['/path/to/parameter:1']:
  if i.split(':', 1)[0] in params:
    i = i.split(':', 1)[0]
    print(params[i])

PARAMETER VALUE IS PRINTED HERE
```

See Above:
```paste below

```

### DIFF
--- a/plugins/lookup/aws_ssm.py
+++ b/plugins/lookup/aws_ssm.py
@@ -223,6 +223,7 @@ class LookupModule(LookupBase):
                                                     tag_value_key_name="Value")
             for i in terms:
                 if i.split(':', 1)[0] in params:
+                    i = i.split(':', 1)[0]
                     ret.append(params[i])
                 elif i in response['InvalidParameters']:
                     ret.append(None)


dginther:
The failing tests above don't seem to be related to my PR in any way :/

dginther:
Opened an issue for this PR: https://github.com/ansible-collections/amazon.aws/issues/389

jillr:
recheck

jillr (commented):
Thanks for this PR @dginther.  The change looks good to me, could you please add a [changelog](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) for the new functionality?

dginther:
> Thanks for this PR @dginther. The change looks good to me, could you please add a [changelog](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to) for the new functionality?

Done! Thanks @jillr

goneri:
recheck

jillr (approved):
recheck

goneri:
@dginther Can you please rebase you branch to address the conflict?

dginther:
> @dginther Can you please rebase you branch to address the conflict?

No, I can't, because the code was _completely changed_ by #378. If I get some free time, I will try to read through the new code and see if it works for this use case, and if it doesn't, will try to fix it.

PR #398, opened 2021-07-07
Install ec2_eni test requirements

jillr:
##### SUMMARY
ansible.netcommon needs the netaddr python package to be installed in the test env, add this to test-requirements.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ec2_eni


jillr:
I'm bad at git today but once this lands it should unblock #337

PR #487, opened 2021-09-03
ec2_spot_instance_info: add new module for describing spot instance requests

mandar242:
##### SUMMARY

Added a new module that describes the specified Spot Instance requests.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

ec2_spot_instance_info

##### ADDITIONAL INFORMATION

```
Related Documentation:
https://docs.aws.amazon.com/goto/WebAPI/ec2-2016-11-15/DescribeSpotInstanceRequests
```

### DIFF
--- a/plugins/modules/ec2_spot_instance_info.py
+++ b/plugins/modules/ec2_spot_instance_info.py
@@ -53,6 +53,16 @@ EXAMPLES = '''
         values:
           - t3.medium
 
+- name: describe the Spot requests filtered using multiple filters
+  amazon.aws.ec2_spot_instance_info:
+    filters:
+      - name: 'state'
+        values:
+          - active
+      - name: launch.block-device-mapping.device-name
+        values:
+          - /dev/sdb
+
 '''
 
 RETURN = '''
@@ -109,9 +119,11 @@ from ansible_collections.amazon.aws.plugins.module_utils.ec2 import AWSRetry
 from ansible.module_utils.common.dict_transformations import snake_dict_to_camel_dict
 from ansible.module_utils.common.dict_transformations import camel_dict_to_snake_dict
 
+
 def _describe_spot_instance_requests(connection, **params):
     paginator = connection.get_paginator('describe_spot_instance_requests')
     return paginator.paginate(**params).build_full_result()
+
 
 def describe_spot_instance_requests(connection, module):
 
@@ -134,10 +146,12 @@ def describe_spot_instance_requests(connection, module):
     except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
         module.fail_json_aws(e, msg='Failed to describe spot instance requests')
 
-    # spot_request = {'spot_instance_requests': []}
     spot_request = []
     for response_list_item in describe_spot_instance_requests_response:
         spot_request.append(camel_dict_to_snake_dict(response_list_item))
+
+    if len(spot_request) == 0:
+        module.exit_json(changed=changed, msg='No spot requests found for specified options')
 
     module.exit_json(spot_request=spot_request, changed=changed)
 
--- a/plugins/modules/ec2_spot_instance_info.py
+++ b/plugins/modules/ec2_spot_instance_info.py
@@ -167,7 +167,7 @@ def describe_spot_instance_requests(connection, module):
     for response_list_item in describe_spot_instance_requests_response:
       spot_request['spot_instance_requests'].append(camel_dict_to_snake_dict(response_list_item))
 
-    module.exit_json(spot_request=spot_request, changed=changed)
+    module.exit_json(spot_request=spot_request['spot_instance_requests'], changed=changed)
 
 
 def main():
--- a/plugins/modules/ec2_spot_instance_info.py
+++ b/plugins/modules/ec2_spot_instance_info.py
@@ -57,7 +57,6 @@ RETURN = '''
     spot_instance_request_ids:
       - sir-c3cp9jsk
 '''
-
 import time
 import datetime
 
--- a/plugins/modules/ec2_spot_instance_info.py
+++ b/plugins/modules/ec2_spot_instance_info.py
@@ -57,6 +57,8 @@ RETURN = '''
     spot_instance_request_ids:
       - sir-c3cp9jsk
 '''
+
+import q
 import time
 import datetime
 
--- a/plugins/modules/ec2_spot_instance_info.py
+++ b/plugins/modules/ec2_spot_instance_info.py
@@ -13,7 +13,6 @@ EXAMPLES = '''
 
 RETURN = '''
 '''
-
 import time
 import datetime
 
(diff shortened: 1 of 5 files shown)


plugins/modules/ec2_spot_instance_info.py
```diff
@@ -0,0 +1,195 @@
...
+  - Mandar Vijay Kulkarni (@mandar242)
+options:
+  filters:
+    description:
+      - A list of filters to apply.
+      - Each list item is a dict consisting of a filter key and a filter value.
+      - See U(https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeSpotInstanceRequests.html) for possible filters.
+    type: list
+    elements: dict
+  dry_run:
```
jillr, line 26: We don't generally use boto3's dry run features. We should be catching all boto errors to the user already so we're covered for this.
We did use it more historically before we had good Ansible check_mode support so you might see it in a few modules still.

plugins/modules/ec2_spot_instance_info.py
```diff
@@ -0,0 +1,195 @@
...
+      - If required permissions are not present then returns an error response UnauthorizedOperation else DryRunOperation.
+    default: False
+    choices: [ True, False ]
+    type: bool
+  spot_instance_request_ids:
+    description:
+      - One or more Spot Instance request IDs.
+    type: list
+    elements: str
+  next_token:
```
jillr, line 39: Instead of using next_token and max_results we should use a paginator on the `describe_spot_instance_requests()` call in the module.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Paginator.DescribeSpotInstanceRequests

jillr (commented):
Looking good so far @mandar242! A couple of comments about conventions that we usually use in the aws modules.

tests/integration/targets/ec2_spot_instance/tasks/main.yaml
```diff
@@ -230,6 +277,19 @@
         ignore_errors: true
         retries: 5
 
+      - name: Get info about the complex spot instance request created
```
jillr, line 280: In the `always` block we want to use the module a little differently.  We shouldn't really be testing here, this is just for cleanup.  But we would like to make use of the info module to remove the hard-coded variables at lines 275 and 275.
```
            - '{{ create_result.spot_request.spot_instance_request_id }}'
            - '{{ complex_create_result.spot_request.spot_instance_request_id }}'
```
Instead we'd add a `ec2_spot_instance_info` task before the`remove spot instance requests` task that filters on vpc-id `vpc_result.vpc.id`.  The result from that should be every single sir we created in the test, whether that's 1, 2, or 20 of them.  Then in the delete task, we can have a `with_items` to loop through.   This way if the test suite grows in the future, we can be sure we're always deleting everything without having to keep adding static lines to the cleanup task.   That would look something like:

```
      - name: remove spot instance requests
        ec2_spot_instance:
          spot_instance_request_ids:
            - '{{ item.spot_instance_request_id }}'
          state: 'absent'
        ignore_errors: true
        retries: 5
        with_items: '{{ var_you_registered_in_info_task.spot_request }}'
```
There's a pretty good example here https://github.com/ansible-collections/amazon.aws/blob/main/tests/integration/targets/ec2_instance/roles/ec2_instance/tasks/env_cleanup.yml#L10-L24

plugins/modules/ec2_spot_instance_info.py
```diff
@@ -0,0 +1,178 @@
...
+      - sir-87654321
+    filters:
+      - name: 'launch.instance-type'
+        values:
+          - t3.medium
+
+- name: describe the Spot requests filtered using multiple filters
+  amazon.aws.ec2_spot_instance_info:
+    filters:
+      - name: 'state'
```
jillr, line 59: We simplify filters for users so that they don't have to put `name` and `values` as keys.  `module_utils.ec2` has a helper `ansible_dict_to_boto3_filter_list` that will do this for you.  That way users can just say:
```
filters:
  - state: active
```

tests/integration/targets/ec2_spot_instance/tasks/main.yaml
```diff
@@ -145,6 +157,35 @@
...
+    register: complex_spot_instance_info_result
+
+  - name: Assert that the complex spot request created is open/active and correct keys are set
+    assert:
+      that:
+        - complex_spot_instance_info_result.spot_request[0].state in ['open', 'active']
+        - complex_create_result.spot_request.spot_price == "0.002000"
+        - complex_create_result.spot_request.launch_specification.block_device_mappings[0].ebs.volume_size == 5
+        - complex_create_result.spot_request.launch_specification.block_device_mappings[0].ebs.volume_type == "gp3"
+        - complex_create_result.spot_request.launch_specification.network_interfaces[0].delete_on_termination == True
```
jillr, line 173: For these we want to compare the info_result to the create_result to ensure the two modules have consistent results. So these would be:
```suggestion
        - complex_create_info_result.spot_request.spot_price == complex_create_result.spot_request.spot_price
        - complex_create_info_result.spot_request.launch_specification.block_device_mappings[0].ebs.volume_size == complex_create_result.spot_request.launch_specification.block_device_mappings[0].ebs.volume_size
        - complex_create_info_result.spot_request.launch_specification.block_device_mappings[0].ebs.volume_type == complex_create_result.spot_request.launch_specification.block_device_mappings[0].ebs.volume_type
        - complex_create_info_result.spot_request.launch_specification.network_interfaces[0].delete_on_termination == complex_create_result.spot_request.launch_specification.network_interfaces[0].delete_on_termination
```

In this specific case though, that gets really long and hard to read. So I would suggest using a task var for all of the launch spec assertions like the example at line 156-158.  Which could make this block instead be something like:

```suggestion
        - complex_create_info_result.spot_request.spot_price == complex_create_result.spot_request.spot_price 
        - info_launch_spec.block_device_mappings[0].ebs.volume_size == create_launch_spec.block_device_mappings[0].ebs.volume_size
        - info_launch_spec.block_device_mappings[0].ebs.volume_type == create_launch_spec.block_device_mappings[0].ebs.volume_type
        - info_launch_spec.network_interfaces[0].delete_on_termination == create_launch_spec.network_interfaces[0].delete_on_termination
    vars:
      create_launch_spec: "{{ complex_create_result.spot_request.launch_specification }}"
      info_launch_spec: "{{ complex_create_info_result.spot_request.launch_specification }}"

```
(I'm reasonably confident my indentation is correct on that one, but it's hard to tell for sure in the github comment pop-up.  :))

tests/integration/targets/ec2_spot_instance/tasks/main.yaml
```diff
@@ -145,6 +157,35 @@
...
+
+  - name: Get info about the created spot instance requests and filter result based on provided filters
+    ec2_spot_instance_info:
+      spot_instance_request_ids:
+        - '{{ create_result.spot_request.spot_instance_request_id }}'
+        - '{{ complex_create_result.spot_request.spot_instance_request_id }}'
+      filters:
+        launch.block-device-mapping.device-name: /dev/sdb
```
jillr, line 181: If there's more than one copy of this test running at the same time in CI this can return multiple results (for example, 2 PRs for the same module that get checked at the same time). We should also filter on the unique tag we talked about adding.

tests/integration/targets/ec2_spot_instance/tasks/main.yaml
```diff
@@ -221,14 +271,26 @@
...
           spot_instance_request_ids:
-            - '{{ create_result.spot_request.spot_instance_request_id }}'
-            - '{{ complex_create_result.spot_request.spot_instance_request_id }}'
+            - '{{ item.spot_instance_request_id }}'
           state: 'absent'
         ignore_errors: true
         retries: 5
+        with_items: "{{ spot_request_list.spot_request }}"
+
+      - name: Assert that the created spot instance requests are removed
```
jillr, line 289: Last small thing then this is ready to merge!

We don't actually want to have this assertion.  It happens to be passing because deleting the instances has the effect of cancelling the spot requests, but this doesn't actually test the result of the removal task because the var is coming from the info task.   We generally don't need assertions in the cleanup block so we can just remove this task.

jillr (approved):
LGTM - thanks for all your work on this @mandar242!

jillr:
EC2 instances in us-east-1 are taking 14+ minutes to reach running, there are likely other delays happening in-region as well.  We'll have to try this (and several other PRs) again tomorrow.

Issue #462, opened 2021-05-05
ec2_instance keeps default value even if set in launch template 


emanuele-leopardi:
##### SUMMARY
It looks like there are still some default values that are not taken into account by this module even if passing a launch template

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

ec2_instance
##### ANSIBLE VERSION

```paste below
ansible 2.10.8
  config file = /Users/ema/Documents/bin/kl_infrastructure/ansible/bda-appquality/ansible.cfg
  configured module search path = ['/Users/ema/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.0 (default, Oct 27 2020, 14:15:17) [Clang 12.0.0 (clang-1200.0.32.21)]
```

##### CONFIGURATION

```paste below
ANSIBLE_SSH_ARGS() = -o ConnectionAttempts=10 -o ControlMaster=auto -o ControlPersist=10m -o StrictHostKeyCh
ANSIBLE_SSH_RETRIES() = 5
DEFAULT_FORCE_HANDLERS() = True
DEFAULT_STRATEGY() = linear
DEFAULT_TIMEOUT() = 120
HOST_KEY_CHECKING() = False
RETRY_FILES_ENABLED() = False
```

##### OS / ENVIRONMENT

MacOS 11.2.3 
##### STEPS TO REPRODUCE

First of all let's define some variables
```yaml
instance:
    name: kl-{{ proj_name }}-backend
    subnet: kl-{{ proj_name }}-db-a
    ami_id: ami-fooobarrrbaaaz
    instance_type: t3.micro
    state: present
    termination_protection: false
    region: "eu-south-1" #Milan
    security_groups:
      - kl-{{ proj_name }}-backend-sg
    network:
      assign_public_ip: true
      delete_on_termination: true
      description: kl-{{ proj_name }}-backend
    env: lab
```
then we use this to create a nice ec2_launch_template

```yaml
- name: Create | Update Launch Template
  ec2_launch_template:
    state: "{{ instance.state | default('present')}}"
    template_name: "{{ instance.name | default(omit) }}-launch-template"
    image_id: "{{ instance.ami_id }}"
    instance_type: "{{ instance.instance_type }}"
    key_name: "{{ instance.key | default('ssh_key') }}"
[...]

Issue #451, opened 2021-08-10
 Cannot use launch_template name in ec2_instance module


markuman:
Two issues (TL;DR)

1.  Cannot use launch_template name in ec2_instance module
    * broken condition
2.  You cannot use `launch_template` paremeter with other `ec2_instance` parameters in the same task (e.g. `instance_type`).
    * that's a general boto3 limitation
    * default values of `ec2_instance` is making this even more diffcult to handle

### Summary

Origin Issue: https://github.com/ansible/ansible/issues/51941  
First trial of fix: https://github.com/ansible-collections/community.aws/pull/111  
Broken condition still exists: https://github.com/ansible-collections/amazon.aws/blob/main/plugins/modules/ec2_instance.py#L1165

However. It's not that easy.  

Imo, LaunchTemplate parameters are not changeable in the context of `run_instance`.  
You must read-out the launch_template parameters and throw them aginst `run_instance` to pretend they came from `module.params`.

`"Failed to create new EC2 instance: An error occurred (Unsupported) when calling the RunInstances operation: The requested configuration is currently not supported. Please check the documentation for supported configurations."`

One issues are the default values the modules applies.  
`{'BlockDeviceMappings': [], 'ClientToken': '37d405af374948e2bd427ff4a7a7fc04', 'InstanceType': 't2.micro', 'LaunchTemplate': {'LaunchTemplateName': 'ubuntu-20.04'}, 'MaxCount': 1, 'MinCount': 1, 'NetworkInterfaces': [{'DeviceIndex': 0, 'SubnetId': 'subnet-3e5fe743'}], 'TagSpecifications': [{'ResourceType': 'volume', 'Tags': [{'Key': 'Name', 'Value': 'lttest01'}]}, {'ResourceType': 'instance', 'Tags': [{'Key': 'Name', 'Value': 'lttest01'}]}]}`

This happen when you run `ec2_instance` without any parameters, just with a `LaunchTemplateName`.
So the module applies
* `BlockDeviceMappings: []`
* `InstanceType: `t2.micro`

basically when I run 

```yml
 - name: test launchtemplate usage
      ec2_instance:
        wait: yes
        name: "{{ name }}"
        instance_type: c5a.large
        launch_template:
          name: ubuntu-20.04
        vpc_subnet_id: subnet-3e5fe743
```

I expect that all values are taken from the launch_template, only the requested will be overwritten (`instance_type` and `vpc_subnet_id`.)

but because of the  default values, when I request

[...]

PR #587, opened 2021-12-09
ec2_instance: Fix launch template condition, handle launch template - default value for instance_type

mandar242:
##### SUMMARY

The ```launch_template``` option in ```ec2_instance``` has a broken condition.
Also the ```launch_template``` option defaults the ```instance_type``` to ```t2.micro``` if not specified and ignores the ```instance_type``` specified in the ```launch_template``` as said in the issue #451.

Fixes 
- #451
- #462 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

```ec2_instance```

##### ADDITIONAL INFORMATION

The change does not break existing functionality as tested by the integration test run locally.
Related to the condition fix in community.aws: https://github.com/ansible-collections/community.aws/pull/111

Ran the following playbook to verify the change.
```
# create a launch template called "test-launch-template" 
    - name: create test launch template
      community.aws.ec2_launch_template:
        template_name: test-launch-template
        image_id: ami-002068ed284xxxxxx
        instance_type: t3a.small
        network_interfaces:
          - associate_public_ip_address: no
            delete_on_termination: yes
            device_index: 0
            groups:
              - sg-xxxxxxxxxxxxxxxxxx
            subnet_id: subnet-xxxxxxxxxxxxxxxxxx
        region: us-east-2
        block_device_mappings:
          - device_name: /dev/sdb
            ebs:
              volume_size: 5
              volume_type: gp3
              delete_on_termination: true
              encrypted: yes
          - device_name: /dev/sdc
            ebs:
              volume_size: 2
              volume_type: gp2
              delete_on_termination: true
              encrypted: no
        tags:
          ssome: tags

# launch a ec2 instance using launch template created earlier - launches t3a.small instance as expected
    - name: test launch template usage
      ec2_instance:
        wait: yes
        name: "test-instance-mk-t3a.small"
        launch_template:
          name: test-launch-template
        vpc_subnet_id: subnet-xxxxxxxxxxxxxxxxxx

# launch ec2 instance using launch template created earlier - override instance type to be launch to t3.xlarge
    - name: test launch template usage - override instance type
      ec2_instance:
        wait: yes
        name: "test-instance-mk-t3.xlarge"
        instance_type: t3.xlarge
        launch_template:
          name: test-launch-template
        vpc_subnet_id: subnet-xxxxxxxxxxxxxxxxxx
```

### DIFF
--- a/changelogs/fragments/587-ec2_instance-default-instance-type-launch-template.yml
+++ b/changelogs/fragments/587-ec2_instance-default-instance-type-launch-template.yml
@@ -1,3 +1,6 @@
 bugfixes:
   - ec2_instance - Add a condition to handle default ```instance_type``` value
     for fix breaking on instance creation with launch template (https://github.com/ansible-collections/amazon.aws/pull/587).
+deprecated_features:
+  - ec2_instance - The default value for ```instance_type``` has been deprecated,
+    in the future release you must set an instance_type or a launch_template".
--- a/changelogs/fragments/587-ec2_instance-default-instance-type-launch-template.yml
+++ b/changelogs/fragments/587-ec2_instance-default-instance-type-launch-template.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - ec2_instance - Add a condition to handle default ```instance_type``` value
+    for fix breaking on instance creation with launch template (https://github.com/ansible-collections/amazon.aws/pull/587).
--- a/plugins/modules/ec2_instance.py
+++ b/plugins/modules/ec2_instance.py
@@ -1972,6 +1972,10 @@ def main():
         ],
         supports_check_mode=True
     )
+
+    if not module.params.get('instance_type') and not module.params.get('launch_template'):
+        module.deprecate("Default value instance_type has been deprecated, in the future you must set an instance_type or a launch_template",
+                         date='2023-01-01', collection_name='amazon.aws')
     result = dict()
 
     if module.params.get('network'):
--- a/plugins/modules/ec2_instance.py
+++ b/plugins/modules/ec2_instance.py
@@ -54,7 +54,7 @@ options:
     description:
       - Instance type to use for the instance, see U(https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html)
         Only required when instance is not already present.
-    default: t2.micro
+      - If not specified, t2.micro will be used.
     type: str
   count:
     description:
@@ -1232,7 +1232,7 @@ def build_top_level_options(params):
 
     if params.get('launch_template') is not None:
         spec['LaunchTemplate'] = {}
-        if not params.get('launch_template').get('id') or params.get('launch_template').get('name'):
+        if not params.get('launch_template').get('id') and not params.get('launch_template').get('name'):
             module.fail_json(msg="Could not create instance with launch template. Either launch_template.name or launch_template.id parameters are required")
 
         if params.get('launch_template').get('id') is not None:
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by mandar242** [comment]
>
Hi @markuman, can you take please a look at this for #451, this change would 
1. Fix the broken condition [here](https://github.com/ansible-collections/amazon.aws/blob/main/plugins/modules/ec2_instance.py#L1235) for ```launch_template``` as fixed earlier by https://github.com/ansible-collections/community.aws/pull/111
2. Handle the ```default``` value for ```instance_type``` to not force ```t2.micro```  in use cases for ```launch_template```.

If this approach looks like a possible solution, I can work for adding few integration tests for the change.

File: plugins/modules/ec2_instance.py [hunk 1]
```diff
@@ -54,7 +54,6 @@
     description:
       - Instance type to use for the instance, see U(https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html)
         Only required when instance is not already present.
-    default: t2.micro
```
jillr, line 57: Removing the actual module default value is probably our best option, but we should add a note here that if no option is specified, t2.micro will be used (so it isn't a surprise if someone just forgets to specify or something).  We should also add a deprecation warning to the code about that the default value is going to go away and in the future folks will always need to specify either an instance_type, or an LT.  Then we can actually deprecate it and avoid this entirely in the future!

markuman (commented):
A changelog fragment is missing.

mandar242:
recheck

mandar242:
recheck

emanuele-leopardi (approved):
LGTM, in fact I think it doesn't make sense to have a default here. IMO, I'd rather have a failing playbook just because I forgot to pick an instance type, instead of a default value :)

changelogs/fragments/587-ec2_instance-default-instance-type-launch-template.yml
```diff
@@ -0,0 +1,3 @@
+bugfixes:
+  - ec2_instance - Add a condition to handle default ```instance_type``` value
+    for fix breaking on instance creation with launch template (https://github.com/ansible-collections/amazon.aws/pull/587).
```
jillr, line 3: Could you please note the deprecation of the default value in the changelog as well?  Otherwise this should be good once the sanity ignore for 2.9 is added - thanks!

jillr (approved):

plugins/modules/ec2_instance.py
```diff
@@ -1967,6 +1972,10 @@ def main():
         ],
         supports_check_mode=True
     )
+
+    if not module.params.get('instance_type') and not module.params.get('launch_template'):
+        module.deprecate("Default value instance_type has been deprecated, in the future you must set an instance_type or a launch_template",
+                         date='2023-01-01', collection_name='amazon.aws')
```
alinabuzachis, line 1978: @jillr Just an info, do we use date or collection version for deprecations?

@mandar242 Other than that, it looks good to me. Thank you for taking time to work on this.

alinabuzachis (approved):

plugins/modules/ec2_instance.py
jillr: @alinabuzachis date, we don't expect to always have the same cadence of major collection releases

jillr:
recheck

jillr:
recheck

jillr:
@mandar242 you can add the gate label to this PR when the sanity/ignores-2.9.txt file is updated and the tests are passing

alinabuzachis:
recheck

PR #615, opened 2022-01-14
Add the ability to disable the validate_bucket_name check

### LINKED ISSUE #593: Bucket name validation is too eager and should be skippable
### Summary

With the implementation of https://github.com/ansible-collections/amazon.aws/pull/341 all bucket names must adhere to Amazon S3's naming rules. While those rules are generally followed by alternative S3 implementations, they are not a given and can sometimes be disabled (at least in the case of Ceph):

> The above constraints are relaxed if the option ‘rgw_relaxed_s3_bucket_names’ is set to true except that the bucket names must still be unique, cannot be formatted as IP address and can contain letters, numbers, periods, dashes and underscores for up to 255 characters long.

https://docs.ceph.com/en/latest/radosgw/s3/bucketops/

Furthermore, the way this works is that the bucket name is checked with every `s3_bucket` operation. So those rules don't just apply to creating new buckets, but they apply to a simple `--check --diff` run as well. At the very least, this is too eager in my estimation. If at all those rules should be applied, it should happen only when necessary.

I would like to change this in one of two ways:

1. Add an option to skip validation that is off by default (e.g. `validate_bucket_names: no (default yes)`).
2. A more fine-grained implementation that only checks the bucket name before creating a bucket, not when updating/reading it.

Those approaches could of course be combined. I'm happy to provide a pull request myself, if there are no objections to my proposal.

### Issue Type

Feature Idea

### Component Name

s3_bucket

### Additional Information

_No response_

### Code of Conduct

href:
The AWS bucket naming rules are not necessarily enforced on all S3 implementations. Where the rules are not enforced, the validation can now be disabled (it is still on by default).

Resolves #593.

### DIFF
--- a/changelogs/fragments/615-s3-validate_bucket_name.yml
+++ b/changelogs/fragments/615-s3-validate_bucket_name.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - aws_s3 - Add `validate_bucket_name` option, to control bucket name validation.
+  - s3_bucket - Add `validate_bucket_name` option, to control bucket name validation.
--- a/plugins/modules/aws_s3.py
+++ b/plugins/modules/aws_s3.py
@@ -197,6 +197,14 @@ options:
         type: str
         description:
         - version ID of the source object.
+  validate_bucket_name:
+    description:
+      - Whether the bucket name should be validated to conform to AWS S3 naming rules.
+      - On by default, this may be disabled for S3 backends that do not enforce these rules.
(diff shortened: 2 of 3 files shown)


jillr (commented):
Thanks for this change @href, could you please add a [changelog fragment](https://github.com/ansible-community/antsibull-changelog/blob/main/docs/changelogs.rst) to tell users about the new option?
The code lgtm

href:
I added a fragment, I hope it's adequate.

jillr (approved):

PR #625, opened 2022-01-20
Add waiter config for rds_instance_available

marknet15:
##### SUMMARY
Currently rarely from time to time on RDS instance creation with `community.aws.rds_instance` I see the following error:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: botocore.exceptions.WaiterError: Waiter DBInstanceAvailable failed: Waiter encountered a terminal failure state
fatal: [vpc_security_groups]: FAILED! => {"boto3_version": "1.16.0", "botocore_version": "1.19.0", "changed": false, "msg": "Error while waiting for DB instance ansible-test-44f5e4d5a14a to be available: Waiter DBInstanceAvailable failed: Waiter encountered a terminal failure state", "resource_actions": ["rds:DescribeDBInstances", "rds:ListTagsForResource"]}
```

This can be re-produced intermittently by running the `community.aws.rds_instance` integration tests.

I think what is missing is the extra waiter configuration.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
waiters.py

##### ADDITIONAL INFORMATION
The `rds_instance` tests are slow and currently disabled but can be run with:
```
ansible-test integration --docker centos8 -v rds_instance --allow-disabled
```

### DIFF
--- a/plugins/module_utils/rds.py
+++ b/plugins/module_utils/rds.py
@@ -146,6 +146,7 @@ def wait_for_instance_status(client, module, db_instance_id, waiter_name):
     waiter_expected_status = {
         'db_instance_deleted': 'deleted',
         'db_instance_stopped': 'stopped',
+        'db_instance_available': 'available',
     }
     expected_status = waiter_expected_status.get(waiter_name, 'available')
     if expected_status == 'available':
@@ -154,6 +155,7 @@ def wait_for_instance_status(client, module, db_instance_id, waiter_name):
         extra_retry_codes = []
     for attempt_to_wait in range(0, 10):
         try:
+            sleep(10) # This seems to help limit failures from the waiter itself
             wait(client, db_instance_id, waiter_name, extra_retry_codes)
             break
         except WaiterError as e:
--- a/plugins/module_utils/rds.py
+++ b/plugins/module_utils/rds.py
@@ -155,8 +155,6 @@ def wait_for_instance_status(client, module, db_instance_id, waiter_name):
         extra_retry_codes = []
     for attempt_to_wait in range(0, 10):
         try:
-            # This seems to help limit failures from the waiter itself
-            sleep(10)
             wait(client, db_instance_id, waiter_name, extra_retry_codes)
             break
         except WaiterError as e:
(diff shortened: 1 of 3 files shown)


alinabuzachis:
@marknet15 Thank you for working on this. Would you be interested in having a look at the rds_instance integration tests? At the moment they have been disabled. In order to test this change, you'd need to open a PR in community.aws where you remove disabled from aliases and in the PR's description add Depends-On: https://github.com/ansible-collections/amazon.aws/pull/625

marknet15:
@alinabuzachis yep sure 👍🏻  I'm currently fixing them up again, to make a small fix elsewhere in rds_instance for IOPs changes

plugins/module_utils/rds.py
```diff
@@ -154,6 +155,8 @@ def wait(client, db_instance_id, waiter_name, extra_retry_codes):
         extra_retry_codes = []
     for attempt_to_wait in range(0, 10):
         try:
+            # This seems to help limit failures from the waiter itself
+            sleep(10)
```
jillr, line 159: If at all possible it would be nice to avoid this; RDS is slow enough on its own!  Would you be able to try debugging what boto is doing when this waiter error happens, since it sounds like you can reproduce it fairly often?  It would be interesting to see if there's some other failure we could capture in the waiter that's getting hidden by something else.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/boto3.html
I've used it on the past like this:
```
 # DEBUG is noisy. Reading it on the console sucks, so use a file
 import logging
 logging.basicConfig(filename='boto.log')
 # You could use name=botocore, or name=boto3. 99.9% of calls will happen in botocore,
 # but leaving this empty captures everything - just in case
     boto3.set_stream_logger(name='', logging=logging.DEBUG)
     paginator = client.get_paginator('list_resource_record_sets')
     record_sets = paginator.paginate(**params).build_full_result()['ResourceRecordSets']
     return {
```

marknet15: Yep I'm not even 100% sure it was making any difference, I'll remove it ! The tests seem better anyway with this slight adjustment, I think there were some other test specific issues.

And thanks for the example for troubleshooting further 👍🏻

marknet15:
recheck

marknet15:
recheck

alinabuzachis:
recheck

marknet15:
I think this PR is no longer needed, I have a feeling most of the issues in rds_instance were integration tests related I've tried lots of testing locally with and without the change here and it didn't seem to be needed. Closing for now. Tests all pass on the rds_instance PR

PR #776, opened 2022-04-18
rds module_util - add retry_code to get_rds_method_attributes and add functionality for copying snapshot

jatorcasso:
##### SUMMARY
- catch InvalidDBSecurityGroupStateFault when modifying db
- add retry_codes to return data from get_rds_method_attribute, and add unit tests to cover
- add functionality for `copy_db_snapshot` 
- refactor to utilize the return values of get_rds_method_attribute

##### ISSUE TYPE
- Feature Pull Request
- Bugfix Pull Request

##### COMPONENT NAME
rds

##### ADDITIONAL INFORMATION
- Ran into the InvalidDBSecurityGroupStateFault error running integration tests for community.aws.rds_instance

There is quite a bit of refactoring here so I wanted to explain why I made the changes I did:
- there were many instances of calling get_rds_method_attribute to retrieve certain values (waiter, retry_codes, which RDS resource), so I tried to remove the duplicate calls and use the return values already retrieved from the initial call.
- split wait_for_snapshot_status into wait_for_instance_snapshot_status and wait_for_cluster_snapshot_status, as they have different identifiers (they are also broken up in get_final_identifier)

The initial goal here was simply to add retry_codes as a return value of get_rds_method_attribute and add `copy_db_snapshot` as a valid method, but working through that uncovered some bugs and just general redundancies, so I felt these were necessary changes.

### DIFF
--- a/plugins/module_utils/rds.py
+++ b/plugins/module_utils/rds.py
@@ -21,7 +21,7 @@ from .ec2 import boto3_tag_list_to_ansible_dict
 from .ec2 import compare_aws_tags
 from .waiters import get_waiter
 
-Boto3ClientMethod = namedtuple('Boto3ClientMethod', ['name', 'waiter', 'operation_description', 'cluster', 'instance', 'snapshot', 'retry_codes'])
+Boto3ClientMethod = namedtuple('Boto3ClientMethod', ['name', 'waiter', 'operation_description', 'resource', 'retry_codes'])
 # Whitelist boto3 client methods for cluster and instance resources
 cluster_method_names = [
     'create_db_cluster', 'restore_db_cluster_from_snapshot', 'restore_db_cluster_from_s3',
@@ -36,26 +36,43 @@ instance_method_names = [
     'remove_role_from_db_instance'
 ]
 
-snapshot_cluster_method_names = [
+cluster_snapshot_method_names = [
     'create_db_cluster_snapshot', 'delete_db_cluster_snapshot', 'add_tags_to_resource', 'remove_tags_from_resource',
     'list_tags_for_resource'
 ]
 
-snapshot_instance_method_names = [
+instance_snapshot_method_names = [
     'create_db_snapshot', 'delete_db_snapshot', 'add_tags_to_resource', 'remove_tags_from_resource',
     'copy_db_snapshot', 'list_tags_for_resource'
 ]
 
 
 def get_rds_method_attribute(method_name, module):
+    '''
+    Returns rds attributes of the specified method.
+
+        Parameters:
+            method_name (str): RDS method to call
+            module: AnsibleAWSModule
+
+        Returns:
+            Boto3ClientMethod (dict):
+                name (str): Name of method
+                waiter (str): Name of waiter associated with given method
+                operation_description (str): Description of method
+                resource (str): Type of resource this method applies to
+                                One of ['instance', 'cluster', 'instance_snapshot', 'cluster_snapshot']
+                retry_codes (list): List of extra error codes to retry on
+
+        Raises:
+            NotImplementedError if wait is True but no waiter can be found for specified method
+    '''
     waiter = ''
     readable_op = method_name.replace('_', ' ').replace('db', 'DB')
-    cluster = False
-    instance = False
-    snapshot = False
+    resource = ''
     retry_codes = []
     if method_name in cluster_method_names and 'new_db_cluster_identifier' in module.params:
-        cluster = True
+        resource = 'cluster'
         if method_name == 'delete_db_cluster':
             waiter = 'cluster_deleted'
         else:
@@ -66,7 +83,7 @@ def get_rds_method_attribute(method_name, module):
         else:
             retry_codes = ['InvalidDBClusterState']
     elif method_name in instance_method_names and 'new_db_instance_identifier' in module.params:
-        instance = True
+        resource = 'instance'
         if method_name == 'delete_db_instance':
             waiter = 'db_instance_deleted'
         elif method_name == 'stop_db_instance':
@@ -84,8 +101,8 @@ def get_rds_method_attribute(method_name, module):
             retry_codes = ['InvalidDBSnapshotState']
         else:
             retry_codes = ['InvalidDBInstanceState', 'InvalidDBSecurityGroupState']
-    elif method_name in snapshot_cluster_method_names and 'db_cluster_snapshot_identifier' in module.params:
-        snapshot = True
+    elif method_name in cluster_snapshot_method_names and 'db_cluster_snapshot_identifier' in module.params:
+        resource = 'cluster_snapshot'
         if method_name == 'delete_db_cluster_snapshot':
             waiter = 'db_cluster_snapshot_deleted'
             retry_codes = ['InvalidDBClusterSnapshotState']
@@ -96,8 +113,8 @@ def get_rds_method_attribute(method_name, module):
             # Tagging
             waiter = 'db_cluster_snapshot_available'
             retry_codes = ['InvalidDBClusterSnapshotState']
-    elif method_name in snapshot_instance_method_names and 'db_snapshot_identifier' in module.params:
-        snapshot = True
+    elif method_name in instance_snapshot_method_names and 'db_snapshot_identifier' in module.params:
+        resource = 'instance_snapshot'
         if method_name == 'delete_db_snapshot':
             waiter = 'db_snapshot_deleted'
             retry_codes = ['InvalidDBSnapshotState']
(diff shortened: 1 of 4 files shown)


alinabuzachis:
@jatorcasso The CI does not test this change with the other rds modules we have in community.aws. This means, we have to manually test those. Can you please open a PR that tests all the available rds modules in community.aws and Depends-On this one?

plugins/module_utils/rds.py
```diff
@@ -156,21 +180,10 @@ def call_method(client, module, method_name, parameters):
...
-                if wait:
-                    wait_for_status(client, module, module.params['db_instance_identifier'], 'modify_db_instance')
-                result = AWSRetry.jittered_backoff(catch_extra_error_codes=['InvalidDBInstanceState'])(method)(**parameters)
-            elif method_name == 'modify_db_cluster':
-                # check if cluster is in an available state first, if possible
-                if wait:
-                    wait_for_status(client, module, module.params['db_cluster_identifier'], method_name)
-                result = AWSRetry.jittered_backoff(catch_extra_error_codes=['InvalidDBClusterStateFault'])(method)(**parameters)
-            else:
-                result = AWSRetry.jittered_backoff()(method)(**parameters)
```
alinabuzachis, line 173: I'm not sure if we want to delete all these. If we delete them, it seems we don't catch anymore the catch_extra_error_codes because we don't handle those methods inside call_method.

jatorcasso: @alinabuzachis we still do! its just using the get_rds_method_attribute to get the error codes instead of in this method - there are integration tests that check for the error messages

plugins/module_utils/rds.py
```diff
@@ -181,27 +194,22 @@ def call_method(client, module, method_name, parameters):
...
 def wait_for_instance_status(client, module, db_instance_id, waiter_name):
-    def wait(client, db_instance_id, waiter_name, extra_retry_codes):
-        retry = AWSRetry.jittered_backoff(catch_extra_error_codes=extra_retry_codes)
+    def wait(client, db_instance_id, waiter_name):
         try:
             waiter = client.get_waiter(waiter_name)
         except ValueError:
             # using a waiter in module_utils/waiters.py
             waiter = get_waiter(client, waiter_name)
-        waiter.wait(WaiterConfig={'Delay': 60, 'MaxAttempts': 60}, DBInstanceIdentifier=db_instance_id)
```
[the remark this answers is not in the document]
jatorcasso: I thought it was redundant since the waiters already define these values based on the specific waiter

plugins/module_utils/rds.py
alinabuzachis: get_rds_method_attribute does not handle  the methods you deleted here, don't understand how you get those retries when the method will be called by rds_instance and rds_cluster.

jatorcasso:
> @jatorcasso The CI seems not testing this change with the other rds modules we have in community.aws. This means, we have to manually test those. Can you please open a PR that tests all the available rds modules in community.aws and Depends-On this one? We have to ensure these changes don't break the other rds modules.

@alinabuzachis  https://github.com/ansible-collections/community.aws/pull/1081

alinabuzachis:
> > @jatorcasso The CI seems not testing this change with the other rds modules we have in community.aws. This means, we have to manually test those. Can you please open a PR that tests all the available rds modules in community.aws and Depends-On this one? We have to ensure these changes don't break the other rds modules.
> 
> @alinabuzachis [ansible-collections/community.aws#1081](https://github.com/ansible-collections/community.aws/pull/1081)

It seems in community.aws it's only run fir rds_instance. You should also open a PR where you only add a string for all the rds modules that are excluded and use module_utils.

jatorcasso:
> @jatorcasso The CI seems not testing this change with the other rds modules we have in community.aws. This means, we have to manually test those. Can you please open a PR that tests all the available rds modules in community.aws and Depends-On this one? We have to ensure these changes don't break the other rds modules.

It runs them all! integration-community.aws-2 runs rds_instance and integration-community.aws-3 runs rds_cluster rds_option_group and rds_subnet_group

jatorcasso:
recheck

jatorcasso:
recheck

jatorcasso:
recheck

jatorcasso:
recheck

plugins/module_utils/rds.py
```diff
@@ -156,21 +180,10 @@ def call_method(client, module, method_name, parameters):
...
     if not module.check_mode:
         wait = module.params.get('wait')
-        # TODO: stabilize by adding get_rds_method_attribute(method_name).extra_retry_codes
+        retry_codes = get_rds_method_attribute(method_name, module).retry_codes
         method = getattr(client, method_name)
         try:
-            if method_name in ['modify_db_instance', 'promote_read_replica']:
-                # check if instance is in an available state first, if possible
-                if wait:
-                    wait_for_status(client, module, module.params['db_instance_identifier'], 'modify_db_instance')
```
jatorcasso, line 165: removed this wait prior to the API call since the `retry_codes` takes care of instances where the resource isn't yet available

jatorcasso:
Recheck

plugins/module_utils/rds.py
```diff
@@ -156,52 +200,39 @@ def call_method(client, module, method_name, parameters):
...
-            else:
-                result = AWSRetry.jittered_backoff()(method)(**parameters)
+            result = AWSRetry.jittered_backoff(catch_extra_error_codes=retry_codes)(method)(**parameters)
         except (BotoCoreError, ClientError) as e:
             changed = handle_errors(module, e, method_name, parameters)
 
         if wait and changed:
             identifier = get_final_identifier(method_name, module)
-            wait_for_status(client, module, identifier, method_name)
+            wait_for_status(client, module, resource, identifier, waiter)
```
alinabuzachis, line 215: I guess you can directly pass `rds_method_attributes` as parameter instead of resource and waiter.

plugins/module_utils/rds.py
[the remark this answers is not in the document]
alinabuzachis: Not sure if this overwrites waiter's config defined in waiters.py. Was just wondering if this was a waiter specific configuration for rds instance only.

plugins/module_utils/rds.py
jatorcasso: @alinabuzachis I suppose - but then id have to unpack them right after anyway so isn't it more efficient to pass in the way it is now?

plugins/module_utils/rds.py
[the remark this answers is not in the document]
jatorcasso: I reverted it, no real reason to change it so ill just keep it as is

plugins/module_utils/rds.py
alinabuzachis: @jatorcasso Mine was only a suggestion. In theory, those will be unpacked only when a waiter will be needed, otherwise, they won't (because useless). This was to also reduce the number of params passed.

jatorcasso: I was just curious haha. `wait_for_status` only gets called when wait is True, so we'll always need the waiter in that case

alinabuzachis: Exactly, only in that case, but these are very small neats. However, I reviewed again and it looks good  for me.

alinabuzachis (approved):

plugins/module_utils/rds.py
alinabuzachis: @jatorcasso Thank you.

jatorcasso: thanks!

plugins/module_utils/rds.py
```diff
@@ -195,13 +230,9 @@ def wait(client, db_instance_id, waiter_name, extra_retry_codes):
         'db_instance_stopped': 'stopped',
     }
     expected_status = waiter_expected_status.get(waiter_name, 'available')
-    if expected_status == 'available':
-        extra_retry_codes = ['DBInstanceNotFound']
```
s-hertel, line 199: I don't see this added back anywhere, should it be added in rds_instance as part of https://github.com/ansible-collections/community.aws/pull/1081? (Of course not when the instance is being deleted)

plugins/module_utils/rds.py
```diff
@@ -249,16 +272,28 @@ def wait_for_snapshot_status(client, module, db_snapshot_id, waiter_name):
...
+        if waiter_name == 'db_cluster_snapshot_deleted':
+            msg = "Failed to wait for DB cluster snapshot {0} to be deleted".format(db_snapshot_id)
+        else:
+            msg = "Failed to wait for DB cluster snapshot {0} to be available".format(db_snapshot_id)
+        module.fail_json_aws(e, msg=msg)
+    except (BotoCoreError, ClientError) as e:
+        module.fail_json_aws(e, msg="Failed with an unexpected error while waiting for the DB cluster snapshot {0}".format(db_snapshot_id))
+
+
+def wait_for_status(client, module, resource, identifier, waiter_name):
```
s-hertel, line 288: How do the AWS collections handle breaking changes for module_utils (maybe breaking_changes changelog fragments)? I've created collections that depend on amazon.aws to be able to use the module_utils (though not these), but changing the function signature would mean that the collection depending on it will also need to update their code.

s-hertel (approved):

plugins/module_utils/rds.py
jatorcasso: The `extra_retry_codes` here were never used, so that's why they're not added back :)

PR #798, opened 2022-04-29
tests: ensure we can import_role the targets

goneri:
List `amazon.aws` in the `collections` list to be sure Ansible can
find the modules at the right place.
This allow us to `import_role` the different targets without using
`ansible-test`.


goneri:
recheck

PR #799, opened 2022-04-29
setup_sshkey: drop a dep on Crypto

goneri:
Adjust ec2-fingerprint.py so it use cryptography instead of the
deprecated Crypto library.


tremble (dismissed):
`ERROR: tests/integration/targets/setup_sshkey/files/ec2-fingerprint.py:1:1: unexpected non-module shebang: b'#!/usr/bin/env python3'`

Please also make sure that at least the `ec2_key` test is triggered.

goneri:
recheck

jillr (approved):

goneri:
regate

PR #809, opened 2022-05-11
Use absolute instead of relative imports

nerijus:
This fixes https://github.com/mitogen-hq/mitogen/issues/896. Although ansible without mitogen works without this PR, please consider accepting this PR, because some other files already have absolute imports, for example, `ec2_instance.py`, `ec2_vpc_endpoint.py`, etc.

### DIFF
--- a/plugins/modules/aws_az_info.py
+++ b/plugins/modules/aws_az_info.py
@@ -73,9 +73,9 @@ except ImportError:
 
 from ansible.module_utils.common.dict_transformations import camel_dict_to_snake_dict
 
-from ..module_utils.core import AnsibleAWSModule
-from ..module_utils.ec2 import AWSRetry
-from ..module_utils.ec2 import ansible_dict_to_boto3_filter_list
+from ansible_collections.amazon.aws.plugins.module_utils.core import AnsibleAWSModule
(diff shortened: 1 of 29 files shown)


tremble (approved):
I'm generally good with this change.  However, we should probably update the guidelines if we're going to do this.  (docs/docsite/rst/dev_guidelines.rst)  @nerijus If you're willing to add something that would be good, otherwise we can follow up afterwards as a house-keeping item.

@jillr @alinabuzachis - Any comments from the Cloud Content team side of things?
@markuman @marknet15 - As active community members any thoughts?

Do we want to put this on hold until the next meeting?

marknet15:
I generally prefer absolute imports anyway so this looks all good to me 👍

marknet15 (approved):

markuman (approved):

alinabuzachis (approved):

nerijus:
> I'm generally good with this change. However, we should probably update the guidelines if we're going to do this. (docs/docsite/rst/dev_guidelines.rst) @nerijus If you're willing to add something that would be good, otherwise we can follow up afterwards as a house-keeping item.
I will better leave updating guidelines to somebody else :)

PR #941, opened 2022-07-22
rds - add the engine_mode parameter

alinabuzachis:
Depends-On: https://github.com/ansible-collections/amazon.aws/pull/1185

Add a new `engine_mode` parameter to `rds_cluster` and also adjust the test-suite follow-up to some recent API changes:

- With the aurora engine, `engine_mode` default now on `serverless`.
- We cannot adjust the `endpoint_port` of a `serverless` cluster.
- aurora clusters are now encrypted by default (`storage_encrypted`)

The new `engine_mode` parameter also allows us to re-enable the  Aurora test of the `rds_instance` integration tests.


goneri:
recheck

plugins/module_utils/rds.py
```diff
@@ -185,8 +185,17 @@ def handle_errors(module, exception, method_name, parameters):
         accepted_engines = [
             'aurora', 'aurora-mysql', 'aurora-postgresql'
         ]
+        accepted_engine_modes = [
```
goneri, line 188: Isn't this redundant since we already validate the user inputs with `argument_spec`.

alinabuzachis: I can't remember the exact reason behind it now. I guess we can remove these now https://github.com/ansible-collections/amazon.aws/pull/941/files#diff-93f87b21085d00540e096e681c0d162bcc8575fe1c5995fc4b11623df4c6622dR191-R194 as long as we validate the specification.

plugins/modules/rds_cluster.py
```diff
@@ -171,11 +171,29 @@
...
           - The name of the database engine to be used for this DB cluster. This is required to create a cluster.
+          - The combinaison of I(engine) and I(engine_mode) may not be supported.
+          - "See AWS documentation for details:
+            L(Amazon RDS Documentation,https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html)."
         choices:
           - aurora
           - aurora-mysql
           - aurora-postgresql
+          - mysql
+          - postgres
```
alinabuzachis, line 182: I guess these two can be removed from this PR since they require other parameters we do not support and a specific version of botocore.

goneri: Agreed, and you've got them covered with your own PR.

plugins/module_utils/rds.py
goneri: With `engine`, `engine_mode` and `engine_version` There is a large range of potential combinations. The API returns a clear message and I would prefer to just rely on this. Also, in the future AWS will probably adjust some of these constraints, if we've got a strict check system we will have to release a new collection version.

plugins/modules/rds_cluster.py
```diff
@@ -389,6 +405,23 @@
     username: "{{ username }}"
     cluster_id: "cluster-{{ resource_prefix }}-restored"
     snapshot_identifier: "cluster-{{ resource_prefix }}-snapshot"
+
+- name: Create an Aurora PostgreSQL cluster and attach an intance
+  rds_cluster:
```
alinabuzachis, line 410:
```suggestion
 amazon.aws. rds_cluster:
```

plugins/modules/rds_cluster.py
```diff
@@ -389,6 +405,23 @@
...
+    state: present
+    engine: aurora-postgresql
+    engine_mode: provisioned
+    cluster_id: '{{ cluster_id }}'
+    username: '{{ username }}'
+    password: '{{ password }}'
+
+- name: Attach a new instance to the cluster
+  rds_instance:
```
alinabuzachis, line 419:
```suggestion
  amazon.aws. rds_cluster:
```

plugins/modules/rds_cluster.py
goneri: I would prefer to also to this later in another PR since we need to go through pretty much the whole test-suite anyway.

goneri: Oh, it's the EXAMPLE block, not the test-suite. You're right, let's fix that.

alinabuzachis, line 410:
```suggestion
  amazon.aws.rds_cluster:
```

plugins/modules/rds_cluster.py
goneri: Sorry, I did the change ^^ but I believe I got confused by the fact you own this branch and I ended up doing the push at the wrong location.

PR #977, opened 2022-08-24
new module: s3_object_info

mandar242:
##### SUMMARY

Adding a new module for retrieving s3 object info.

Currently supports following boto APIs
[get_object()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object)
[get_object_acl()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object_acl)
[get_object_attributes()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object_attributes)
[get_object_legal_hold()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object_legal_hold)
[get_object_lock_configuration()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object_lock_configuration)
[get_object_retention()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object_retention)
[get_object_tagging()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object_tagging)

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

s3_object_info
##### ADDITIONAL INFORMATION

Sample Tasks:

```paste below
    - name: Get object info
      amazon.aws.s3_object_info:
        bucket_name: "{{ bucket_name }}"
        object_key: "{{ object_key }}"

    - name: Get object info details
      amazon.aws.s3_object_info:
        bucket_name: "{{ bucket_name }}"
        object_key: "{{ object_key }}"
        object_details:
          object_acl: true
          object_tagging: true
          object_attributes: true
          object_retention: true
        object_attributes:
          - ETag
          - ObjectSize
          - StorageClass
```


plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,570 @@
...
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+module: s3_object_info
+version_added: 5.0.0
+short_description: Gather information about objects in s3
```
tremble, line 13:
```suggestion
short_description: Gather information about objects in S3
```

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,570 @@
...
+description:
+    - Describes the objects in s3.
```
tremble, line 15:
```suggestion
    - Describes objects in S3.
```

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,570 @@
...
+author:
+  - Mandar Vijay Kulkarni (@mandar242)
+options:
+  bucket_name:
+    description:
+      - The bucket name that contains the object.
```
tremble, line 21:
```suggestion
      - The name of the bucket that contains the object.
```

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,570 @@
...
+    required: true
+    type: str
+  object_name:
+    description:
+      - The key of the object.
```
tremble, line 26:
```suggestion
      - The name of the object.
```

tests/integration/targets/s3_object/meta/main.yml
```diff
@@ -1,2 +1,6 @@
 dependencies:
   - setup_remote_tmp_dir
+  # required for s3.get_object_attributes
+  - role: setup_botocore_pip
+    vars:
+      boto3_version: "1.22.0"
```
tremble, line 6:
```suggestion
      boto3_version: "1.22.0"
```

a) Please use botocore versions this is where the API definitions live.
b) I don't see this requirement documented.

tremble (changes requested):

tests/integration/targets/s3_object/tasks/get_object_info.yml
```diff
@@ -0,0 +1,123 @@
...
+        var: info_result
+
+    - assert:
+        that:
+          - info_result is not failed
+          - info_result is not changed
+          - '"object_info" in info_result'
+          - info_result.object_info | length != 0
+
+    - name: Get s3 object info
```
alinabuzachis, line 68: I didn't dig into, but I'm just asking for further info. This is the log of this tasks

```
"object_info": {
2294	2022-08-24 21:30:03.591254 | controller |         "object_acl": {
2295	2022-08-24 21:30:03.591258 | controller |             "msg": "Object ACL not found, are you sure the bucket/object configuration exist?"
2296	2022-08-24 21:30:03.591262 | controller |         },
2297	2022-08-24 21:30:03.591266 | controller |         "object_attributes": {
2298	2022-08-24 21:30:03.591270 | controller |             "msg": "Object attributes not found, are you sure the bucket/object configuration exist?"
2299	2022-08-24 21:30:03.591275 | controller |         },
2300	2022-08-24 21:30:03.591278 | controller |         "object_lock_configuration": {
2301	2022-08-24 21:30:03.591282 | controller |             "msg": "Object lock configuration not found, are you sure the bucket/object configuration exist?"
2302	2022-08-24 21:30:03.591287 | controller |         },
2303	2022-08-24 21:30:03.591291 | controller |         "object_tagging": {
2304	2022-08-24 21:30:03.591294 | controller |             "CamelCase": "SimpleCamelCase",
2305	2022-08-24 21:30:03.591298 | controller |             "Title Case": "Hello Cruel World",
2306	2022-08-24 21:30:03.591302 | controller |             "lowercase spaced": "hello cruel world",
2307	2022-08-24 21:30:03.591306 | controller |             "snake_case": "simple_snake_case"
2308	2022-08-24 21:30:03.591310 | controller |         }
2309	2022-08-24 21:30:03.591314 | controller |     },
```

Since in the log returned by the previous task https://github.com/ansible-collections/amazon.aws/pull/977/files#diff-3b47451f92053737894c5813b0701a668c1ec5f6feb02e78ce3b7b11770f6a2fR53 e_tag had a value ("e_tag": "\"00692ee0f2460b4f4f582f0b0f91a6d9\"",), shouldn't we expect to have it as an attribute as well?

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,570 @@
...
+        if not module.params.get('object_attributes'):
+            module.fail_json(msg='Please provide object_attributes list to retrieve s3 object_attributes.')
+
+    if requested_details:
+        result['object_info'] = get_object_details(connection, module, bucket_name, object_name, requested_details)
+    else:
+        # if specific details are not requested, return object metadata
+        result['object_info'] = get_object(connection, module, bucket_name, object_name)
+
+    module.exit_json(msg="Retrieved s3 object info ", **result)
```
alinabuzachis, line 566: Probably the msg seems useless to me, but I don't have a strong opinion about it.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,570 @@
...
+                                    type: str
+                                    sample: "CanonicalUser"
+                        permission:
+                            description: Specifies the permission given to the grantee.
+                            returned: always
+                            type: str
+                            sample: "FULL CONTROL"
+        object_legal_hold:
+            description: Object's current legal hold status
+            returned: when I(object_legal_hold) is set to I(true) and required configuration is set on bucket.
```
alinabuzachis, line 162: Can you please explain what you mean with "configuration is set on bucket."?

mandar242: yes, so for example if user want to get object lock configuration info, `object_lock_configuration: true`, the `bucket` needs to have `Object Lock Configuration` set for bucket itself.
Otherwise `connection.get_object_legal_hold(**params)` will return boto error as 
```

ClientError('An error occurred (InvalidRequest) when calling the GetObjectLegalHold operation: 
Bucket is missing Object Lock Configuration')

```

One possible way to handle this instead I was thinking, is to just return an empty {} in such case where some configuration for the bucket does not exist, that way I can change documentation for that config to `returned: always`.

what do you think? @alinabuzachis @tremble

tests/integration/targets/s3_object/tasks/get_object_info.yml
```diff
@@ -0,0 +1,123 @@
...
+          CamelCase: "SimpleCamelCase"
+          snake_case: "simple_snake_case"
+
+    - name: List keys
+      amazon.aws.s3_object:
+        bucket: "{{ bucket_name }}-object-info-test"
+        mode: list
+      register: list_keys_result
+
+    - name: Get s3 object info
```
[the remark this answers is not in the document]
mandar242: Added under a new key called `object_metadata`, which will be returned either case where user specifies the object_details they want or even if do not specify object_details they want.
So general output will like

```

"object_info": [
            {
                "object_acl": {
                    "grants": [
                        {
                            "grantee": {
                                "id": "xxxxxxxxxxxxxxxxxxxxxxxxxxxx",
                                "type": "CanonicalUser"
                            },
                            "permission": "FULL_CONTROL"
                        }
                    ],
                    "owner": {
                        "id": "xxxxxxxxxxxxxxxxxxxxxxxxxxxx"
                    }
                },
                "object_attributes": {
                    "e_tag": "xxxxxxxxxxxxxxxxxxxxxxxx",
                    "last_modified": "2022-08-10T01:11:03+00:00",
                    "object_size": 819,
                    "storage_class": "STANDARD",
                    "version_id": "null"
                },
                "object_metadata": {
                    "accept_ranges": "bytes",
                    "content_length": 819,
                    "content_type": "application/x-yaml",
                    "e_tag": "\"xxxxxxxxxxxxxxxxxxx\"",
                    "last_modified": "2022-08-10T01:11:03+00:00",
                    "metadata": {},
                    "tag_count": 2,
                    "version_id": "null"
                },
                "object_tagging": {
                    "a": "1",
                    "b": "2"
                }
            }
        ]

```

tests/integration/targets/s3_object/meta/main.yml
mandar242: a) Fixed
b) not sure if I understand the point correctly, the `s3.get_object_attributes()` was added in `botocore 1.24.7` and `boto3 1.22.0`.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,614 @@
...
+
+
+def get_object_details(connection, module, bucket_name, object_name, requested_facts):
+
+    all_facts = {}
+
+    # Remove non-requested facts
+    requested_facts = {fact: value for fact, value in requested_facts.items() if value is True}
+
+    all_facts['object_metadata'] = get_object(connection, module, bucket_name, object_name)['object_metadata']
```
alinabuzachis, line 518: I won't probably call it `object_metadata`, it is probably misleading, but I do not have a strong opinion about it.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,614 @@
...
+    description:
+      - The name of the bucket that contains the object.
+    required: true
+    type: str
+  object_name:
+    description:
+      - The name of the object.
+    required: false
+    type: str
+  object_attributes:
```
alinabuzachis, line 29: I don't think we need to have two `object_attributes` def We can probably move this one inside object details to overwrite https://github.com/ansible-collections/amazon.aws/pull/977/files#diff-c19ed6d5998f0666b7a49cc21f1da8909f61a4c66f4a8835fd2704ab3542126aR49

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,614 @@
...
+    params = {}
+    params['Bucket'] = bucket_name
+    params['Key'] = object_name
+
+    result = {}
+
+    try:
+        object_info = connection.get_object(**params)
+    except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
+        pass
```
alinabuzachis, line 553: I guess we should fail if some error occurs. We should also handle the case when the specified object is not found.

tests/integration/targets/s3_object/meta/main.yml
tremble: Sorry for the delayed response.

If the module requires a specific version of botocore, then that requirement should be in the module's documentation.  I'm missing where this is called out in the s3_object_info module.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,614 @@
...
+    type: str
+  object_name:
+    description:
+      - The name of the object.
+    required: false
+    type: str
+  object_attributes:
+    description:
+      - The fields/details that should be returned.
+      - Required when I(object_attributes) is C(true) in I(object_details).
```
tremble, line 32: (non-blocking)

I'd be tempted to default this to returning the 5 current attributes.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,614 @@
...
+    import botocore
+except ImportError:
+    pass  # Handled by AnsibleAWSModule
+
+from ansible_collections.amazon.aws.plugins.module_utils.core import AnsibleAWSModule
+from ansible_collections.amazon.aws.plugins.module_utils.ec2 import AWSRetry
+from ansible_collections.amazon.aws.plugins.module_utils.ec2 import camel_dict_to_snake_dict
+from ansible_collections.amazon.aws.plugins.module_utils.ec2 import boto3_tag_list_to_ansible_dict
+
+
```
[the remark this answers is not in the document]
tremble: It would be helpful if this module **did** support listing all the objects in a bucket, that would mean we could deprecate ``state=list`` over in s3_object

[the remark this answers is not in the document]
mandar242: Added support to replicate behavior of `s3_object` `state=list`.

plugins/modules/s3_object_info.py
mandar242: that makes sense, not entirely sure about what could be a good name for the output param here though, considering the data it returns is from https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_object

plugins/modules/s3_object_info.py
mandar242: Renamed to `attributes_list` and moved inside `object_details`.

```
- name: Retrieve detailed S3 object information
  amazon.aws.s3_object_info:
    bucket_name: MyTestBucket
    object_name: MyTestObject
    object_details:
      object_acl: true
      object_attributes: true
      attributes_list:
        - ETag
        - ObjectSize
        - StorageClass

```

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,614 @@
...
+    params = {}
+    params['Bucket'] = bucket_name
+    params['Key'] = object_name
+
+    object_tagging_info = {}
+
+    try:
+        object_tagging_info = connection.get_object_tagging(**params)
+    except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
+        object_tagging_info['msg'] = 'Object tags not found, are you sure the bucket/object tags exist?'
```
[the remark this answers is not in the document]
mandar242: fixed.

tests/integration/targets/s3_object/meta/main.yml
mandar242: @tremble added https://github.com/ansible-collections/amazon.aws/pull/977/files#diff-c19ed6d5998f0666b7a49cc21f1da8909f61a4c66f4a8835fd2704ab3542126aR638-R640

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,658 @@
...
+    bucket_name = module.params.get('bucket_name')
+    object_name = module.params.get('object_name')
+    requested_details = module.params.get('object_details')
+
+    if requested_details and len(requested_details['attributes_list']) != 0:
+        module.require_botocore_at_least('1.24.7', reason='required for s3.get_object_attributes')
+        if not requested_details['attributes_list']:
+            module.fail_json(msg='Please provide attributes_list list to retrieve s3 object_attributes.')
+
+    if requested_details:
```
alinabuzachis, line 643: What is requested_details are specified and no object_name is provided? I would expect to have requested_details for every object in the bucket. Am I wrong?

@jillr @tremble

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,658 @@
...
+    params = {}
+    params['Bucket'] = bucket_name
+
+    result = []
+    list_objects_response = {}
+
+    try:
+        list_objects_response = _list_bucket_objects(connection, **params)
+    except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
+        pass
```
alinabuzachis, line 596: I guess we should fail is something arises here instead.

plugins/modules/s3_object_info.py
alinabuzachis: @mandar242 What happened is the specified object does not exist? What if the S3 bucket does not exist?

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,658 @@
...
+def get_object(connection, bucket_name, object_name):
+    params = {}
+    params['Bucket'] = bucket_name
+    params['Key'] = object_name
+
+    result = {}
+    object_info = {}
+
+    try:
+        object_info = connection.get_object(**params)
```
jillr, line 565: This will download the actual object.  If the object is large this will have a big impact on performance.  client.head_object should provide the same keys without downloading the actual file.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,658 @@
...
+        object_name=dict(type='str'),
+    )
+
+    module = AnsibleAWSModule(
+        argument_spec=argument_spec,
+        supports_check_mode=True,
+    )
+
+    try:
+        connection = module.client('s3', retry_decorator=AWSRetry.jittered_backoff())
```
tremble, line 628: I've been thinking about this.  One problem with using module.client('s3'...) for s3_object_info...

s3_object currently has a number of 'forbidden' states (eg. list), and supports not just AWS, but supports other 'S3-compatible' services such as 'FakeS3' and Ceph/RDB.  Adding s3_object_info which only supports AWS is likely to cause us problems further down the line.

jillr: I have mixed feeling about this.  We have a lot of modules that attempt to support various reimplementations of the S3 API, but we have no ability to test/debug these other services.  Probably suggests that someday this is going to bite us.  For the modules that already implement handling for ceph, DO, etc we shouldn't disable that support lightly but I'm not convinced we should always try to support anything other than AWS for new modules/features.

tremble: The trouble is that to drop the 'forbidden' states from s3_object we need to make sure that s3_object_info supports it.  If s3_object_info doesn't support the same APIs then dropping support for 'list' et al in s3_object means completely dropping support.  For any other module I agree, it's lucky happenstance if it works.  For S3 because we've explicitly said we support them I feel like we're boxed in

jillr: Ah yeah, I forgot we were doing that here.  You're right then, but I don't like it.  :smiley:

plugins/modules/s3_object_info.py
mandar242: @alinabuzachis thank you! good catch, added code modifications required.
The module now will return a list of dicts containing requested details for all the objects in a bucket when object name is not specified.

sample task

```
- name: Retrieve detailed S3 information for all objects in the bucket
  amazon.aws.s3_object_info:
    bucket_name: MyTestBucket
    object_details:
      object_acl: true
      object_attributes: true
      attributes_list:
        - ETag
        - ObjectSize
        - StorageClass
```

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,658 @@
...
+            object_acl=dict(type='bool', default=False),
+            object_legal_hold=dict(type='bool', default=False),
+            object_lock_configuration=dict(type='bool', default=False),
+            object_retention=dict(type='bool', default=False),
+            object_tagging=dict(type='bool', default=False),
+            object_attributes=dict(type='bool', default=False),
+            attributes_list=dict(type='list', elements='str', choices=['ETag', 'Checksum', 'ObjectParts', 'StorageClass', 'ObjectSize']),
+        )),
+        bucket_name=dict(required=True, type='str'),
+        object_name=dict(type='str'),
```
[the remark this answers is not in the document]
mandar242: Not entirely sure on this though, as 

if only `object_name` is specified and `object_details` are not specified module will return object_metadata using `connection.get_object(**params)` [here](https://github.com/ansible-collections/amazon.aws/pull/977/files#diff-c19ed6d5998f0666b7a49cc21f1da8909f61a4c66f4a8835fd2704ab3542126aR655-R658)

while if only `object_details` are specified and no `object_name` is given, module will return info for all the objects in bucket ([added this recently with commit here](https://github.com/ansible-collections/amazon.aws/pull/977/commits/71a506881526549bfb9a0dad16aed450669e86a2))

plugins/modules/s3_object_info.py
mandar242: Fixed.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,658 @@
...
+    except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
+        module.fail_json_aws(e, msg='Failed to connect to AWS')
+
+    result = []
+
+    bucket_name = module.params.get('bucket_name')
+    object_name = module.params.get('object_name')
+    requested_details = module.params.get('object_details')
+
+    if requested_details and len(requested_details['attributes_list']) != 0:
```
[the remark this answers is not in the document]
mandar242: Fixed.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,658 @@
...
+    argument_spec = dict(
+        object_details=dict(type='dict', options=dict(
+            object_acl=dict(type='bool', default=False),
+            object_legal_hold=dict(type='bool', default=False),
+            object_lock_configuration=dict(type='bool', default=False),
+            object_retention=dict(type='bool', default=False),
+            object_tagging=dict(type='bool', default=False),
+            object_attributes=dict(type='bool', default=False),
+            attributes_list=dict(type='list', elements='str', choices=['ETag', 'Checksum', 'ObjectParts', 'StorageClass', 'ObjectSize']),
+        )),
```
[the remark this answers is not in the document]
mandar242: Fixed.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,658 @@
...
+    result = []
+
+    bucket_name = module.params.get('bucket_name')
+    object_name = module.params.get('object_name')
+    requested_details = module.params.get('object_details')
+
+    if requested_details and len(requested_details['attributes_list']) != 0:
+        module.require_botocore_at_least('1.24.7', reason='required for s3.get_object_attributes')
+        if not requested_details['attributes_list']:
+            module.fail_json(msg='Please provide attributes_list list to retrieve s3 object_attributes.')
```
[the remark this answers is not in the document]
mandar242: Fixed.

plugins/modules/s3_object_info.py
mandar242: Fixed. Added check(s) to determine if bucket or a object exists and/or has access permissions.

plugins/modules/s3_object_info.py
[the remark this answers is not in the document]
gravesm: Yeah, with the new changes this can be resolved.

plugins/modules/s3_object_info.py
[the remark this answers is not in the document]
gravesm: I really think the best way to fix this is to use the arg spec as I suggested here, not to manually validate the provided params in the module. With the change you made, the module would go through the process of creating a client, checking if the bucket exists, and then checking if the object exists, only to then fail because of a parameter validation problem. This also means the error message the user gets will look different for this specific case of invalid parameters. Let ansible handle parameter validation. That's its job.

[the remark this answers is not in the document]
mandar242: Added.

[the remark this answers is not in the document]
gravesm: Thanks!

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,701 @@
...
+  returned: when only I(bucket_name) is specified and I(object_name), I(object_details) are not specified.
+  type: list
+  elements: str
+  sample:
+  - prefix1/
+  - prefix1/key1
+  - prefix1/key2
+object_info:
+    description: S3 object details.
+    returned: always
```
alinabuzachis, line 136: Are you sure `object_info` is always returned?

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,701 @@
...
+        default: false
+      object_tagging:
+        description:
+          - Retreive S3 object Tags.
+        required: false
+        type: bool
+        default: false
+      object_attributes:
+        description:
+          - Retreive S3 object attributes.
```
alinabuzachis, line 68: Can you add a note here that this option requires specific botocore version?

mandar242: Fixed.

plugins/modules/s3_object_info.py
mandar242: Fixed.

gravesm (approved):
For my part this LGTM. I'm not sure if @jillr or @tremble have specific suggestions regarding the use of the client and s3 compatible services.

plugins/modules/s3_object_info.py
mandar242: @jillr @tremble I have tried to [add the code here](https://github.com/ansible-collections/amazon.aws/pull/977/commits/7ce7b0e1e678190d3539cfbc44a7e6a3925c1f76) that will support `list_keys` for 's3-compatible' services but would be great help to get guidance on this with my limited experience with such services.
Also not entirely sure how to test the code here? do we have capacity to test the changes with instances in those services?

mandar242: Just realized that s3_object has had recent changes in param specifications with https://github.com/ansible-collections/amazon.aws/pull/994/files#diff-56c6fa0d10ca0b451d1e53e42a9097a3f5c907a785258cc75b4fe93f01ee0612
WIP for pulling changes to `s3_object_info.`

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,811 @@
...
+        description:
+          - The fields/details that should be returned.
+          - Required when I(object_attributes) is C(true) in I(object_details).
+        type: list
+        elements: str
+        choices: ['ETag', 'Checksum', 'ObjectParts', 'StorageClass', 'ObjectSize']
+notes:
+  - Support for I(tags) and I(purge_tags) was added in release 2.0.0.
+  - In release 5.0.0 the I(s3_url) parameter was merged into the I(endpoint_url) parameter,
+    I(s3_url) remains as an alias for I(endpoint_url).
```
alinabuzachis, line 103: Why do we have these notes inn here?

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,811 @@
...
+  endpoint_url:
+    description:
+      - S3 URL endpoint for usage with Ceph, Eucalyptus and fakes3 etc. Otherwise assumes AWS.
+    type: str
+  dualstack:
+    description:
+      - Enables Amazon S3 Dual-Stack Endpoints, allowing S3 communications using both IPv4 and IPv6.
+    type: bool
+    default: false
+  ceph:
```
goneri, line 41: I would just remove this for the same reason. @jillr any opinion?

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,811 @@
...
+'''
+
+EXAMPLES = r'''
+# Note: These examples do not set authentication details, see the AWS Guide for details.
+
+- name: Retrieve a list of objects in S3 bucket
+  amazon.aws.s3_object_info:
+    bucket_name: MyTestBucket
+
+- name: Retrieve a list of objects in Ceph RGW S3
```
goneri, line 118: The same. I don't think it's a good idea to demonstrate how to use the module with a use-case we don't support.

tests/integration/targets/s3_object/tasks/get_object_info.yml
```diff
@@ -0,0 +1,128 @@
...
+          CamelCase: "SimpleCamelCase"
+          snake_case: "simple_snake_case"
+
+    - name: List keys
+      amazon.aws.s3_object_info:
+        bucket_name: "{{ bucket_name }}-object-info-test"
+      register: list_keys_result
+
+    - name: Get s3 object info
+      amazon.aws.s3_object_info:
```
goneri, line 53: There is a lot of redundancy with `s3_object` tests. We end up spending a lot of time just to prepare/teardown the set-up.

Can we just call these `s3_object_info` tasks somewhere during `s3_object` tests and remove this extra test target?

tests/integration/targets/s3_object/meta/main.yml
```diff
@@ -1,2 +1,6 @@
 dependencies:
   - setup_remote_tmp_dir
+  # required for s3.get_object_attributes
```
goneri, line 3: It's a lot of effort just to test a botocore function and we may miss some problem with the current support botocore version. I would prefer we stay we the default botocore version and we don't test `object_attributes`.

goneri (changes requested):

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,811 @@
...
+
+
+DOCUMENTATION = r'''
+---
+module: s3_object_info
+version_added: 5.0.0
+short_description: Gather information about objects in S3
+description:
+    - Describes objects in S3.
+    - Compatible with AWS, DigitalOcean, Ceph, Walrus, FakeS3 and StorageGRID (only supports list_keys currently).
```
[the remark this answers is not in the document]
tremble: See https://github.com/ansible-collections/amazon.aws/pull/977#pullrequestreview-1106205983

We're kinda stuck because the s3_object (aws_s3) module has 'supported' them since the start of this collection, and part of the reason to create s3_object_info is so that we can move the 'info' functionality out of that module.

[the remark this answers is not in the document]
goneri: Thank you @tremble.

plugins/modules/s3_object_info.py
```diff
@@ -0,0 +1,806 @@
...
+# To get S3 connection, in case of dealing with ceph, dualstack, etc.
+def is_fakes3(endpoint_url):
+    """ Return True if endpoint_url has scheme fakes3:// """
+    if endpoint_url is not None:
+        return urlparse(endpoint_url).scheme in ('fakes3', 'fakes3s')
+    else:
+        return False
+
+
+def get_s3_connection(module, aws_connect_kwargs, location, ceph, endpoint_url, sig_4=False):
```
alinabuzachis, line 677: We can keep it as it is for the moment. For the future, we should move this in module_utils instead of introducing redundancy and refactor the other s3 modules as needed.

mandar242: Agreed. Makes sense.
Once this (and other s3 PRs if any) are merged, I can work on refactoring this.

alinabuzachis:
recheck

alinabuzachis (approved):
@mandar242 You should also add s3_object_info here https://github.com/ansible-collections/amazon.aws/blob/main/tests/integration/targets/s3_object/aliases

Other than that, LGTM! Thank you @mandar242

jillr (approved):

mandar242:
regate

alinabuzachis:
regate

PR #981, opened 2022-08-26
tests: use Fedora image from the Fedora org

goneri:
We recently face cases where the search with owner==`amazon` was returning
nothing.

See: https://github.com/ansible-collections/community.aws/pull/1429


goneri:
recheck

PR #986, opened 2022-08-30
inventory/aws_rds: fix a recent regression

goneri:
Put back the `self._read_config_data(path)` call that was removed during
478022695b333043857a6929b350a2a3c07ae567.

Without this call we end up with an empty `aws_rds` in the output if
the inventory file is empty.

To reproduce:
```
cd tests/integration/targets/inventory_aws_rds
ANSIBLE_INVENTORY_ENABLED=amazon.aws.aws_rds ANSIBLE_INVENTORY=test.aws_rds.yml ansible-playbook playbooks/test_invalid_aws_rds_inventory_config.yml -vvvv
```

```
ok: [127.0.0.1] => {
    "groups": {
        "all": [],
        "aws_rds": [],   <--- the unexpected key
        "ungrouped": []
    }
```


goneri:
See: https://github.com/ansible-collections/amazon.aws/pull/987

PR #1002, opened 2022-09-08
recorder: encypt the tarball

goneri:
Reduce the risk of a potential secret leakage with encryption.


goneri:
recheck

goneri:
recheck

goneri:
recheck

tremble (changes requested):
I'm -1 to this.  It makes it impossible to run the tests outside of CI.  I'm good with finding a way to anonymise things, but encrypting the test tarball feels very wrong.

As an alternative it's probably better to leave the recording as text files rather than a tarball, since follks may at least be able to see that they just tried to push sensitive data.

goneri:
@gravesm how do you feel about this?

goneri:
The main advantage I see with a tar.gz are that the modern IDE won't try to index the tarballs. Also the content is supposed to be immutable and a tarball help to avoid any surprise. This is also a way to avoid long diff in Github interface, but I believe they've got a mechanism to turn off diff on some specific files.

gravesm:
@tremble makes a good point. It makes me nervous to just blindly record an AWS session and push it to GH, though. Can we make `record.sh` anonymize anything that looks like an AWS credential, rather than looking for specific credentials?

jillr:
We should step back and do the research task that was identified for this work when it gets prioritized into the backlog so that we can plan for these cases.

goneri:
> We should step back and do the research task that was identified for this work when it gets prioritized into the backlog so that we can plan for these cases.

PR #1015, opened 2022-09-12
tests: add unit-tests for calculate_etag()

goneri:
Add test coverage for `calucalte_etag` that cover the two cases (with or without multipart).


goneri:
recheck

jillr (approved):

PR #1019, opened 2022-09-13
tests: some ec2_instance targets are slower than normal

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1625

Dedicate more time for the execution of some slow `ec2_instance` targets.

See: https://github.com/ansible-collections/amazon.aws/pull/978#issuecomment-1245671362


goneri:
recheck

Issue #950, opened 2022-08-03
aws_ec2 inventory running against the same host multiple times


yamjoepobuda:
### Summary

We recently bumped `amazon.aws` collection to `4.1.0` from `4.0.0` and started experiencing errors due to process locks and other similar errors. We found that the inventory file was being parsed incorrectly. Instead of processing each host once, it appears to be processing each host for every IP address it has, in addition to the usual tag we use to name hosts.

### Issue Type

Bug Report

### Component Name

aws_ec2

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.12.7]
  config file = /etc/ansible/ansible-amer.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.10/dist-packages/ansible
  ansible collection location = /usr/share/ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]
  jinja version = 3.1.2
  libyaml = False
```

### Collection Versions

```console (paste below)
+ ansible-galaxy collection list

# /usr/share/ansible/collections/ansible_collections
Collection            Version
--------------------- -------
amazon.aws            4.1.0  
ansible.netcommon     3.1.0  
ansible.posix         1.4.0  
ansible.utils         2.6.1  
ansible.windows       1.10.0 
chocolatey.chocolatey 1.3.0  
cisco.ios             3.3.0  
cisco.meraki          2.10.1 
commscope.icx         1.0.5  
community.aws         4.0.0  
community.crypto      2.4.0  
community.general     5.4.0  
community.network     4.0.1  
community.windows     1.10.0 
fortinet.fortios      2.1.6  

# /usr/local/lib/python3.10/dist-packages/ansible_collections
Collection                    Version
----------------------------- -------
amazon.aws                    2.3.0  
ansible.netcommon             2.6.1  
ansible.posix                 1.4.0  
ansible.utils                 2.6.1  
ansible.windows               1.10.0 
arista.eos                    3.1.0  
awx.awx                       19.4.0 
azure.azcollection            1.13.0 
check_point.mgmt              2.3.0  
chocolatey.chocolatey         1.2.0  
cisco.aci                     2.2.0  
cisco.asa                     2.1.0  
cisco.dnac                    6.4.0  
cisco.intersight              1.0.19 
[...]

PR #1026, opened 2022-09-15
inventory/aws_ec2: allow multi-entries for one host

goneri:
Depends-On: https://github.com/ansible-collections/amazon.aws/pull/1066

Add an option to allow multiple duplicated entries for one single instance.

Closes: #950

### DIFF
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -751,8 +751,10 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         for host in hosts:
             if allow_duplicated_hosts:
                 hostname_list = self.get_all_hostnames(host, hostnames)
+            elif preferred_hostname := self._get_preferred_hostname(host, hostnames):
+                hostname_list = [preferred_hostname]
             else:
-                hostname_list = [self._get_preferred_hostname(host, hostnames)]
+                continue
 
             host = camel_dict_to_snake_dict(host, ignore_list=['Tags'])
             host['tags'] = boto3_tag_list_to_ansible_dict(host.get('tags', []))
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -71,6 +71,7 @@ options:
         and you want to get all the hostnames that match.
     type: bool
     default: False
+    version_added: 5.0.0
   filters:
     description:
       - A dictionary of filter value pairs.
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -747,6 +747,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             :param hosts: a list of hosts to be added to a group
             :param group: the name of the group to which the hosts belong
             :param hostnames: a list of hostname destination variables in order of preference
+            :param bool allow_duplicated_hosts: allow multiple copy for the same host with a different name
         '''
         for host in hosts:
             if allow_duplicated_hosts:
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -66,7 +66,7 @@ options:
         required: False
   allow_duplicated_hosts:
     description:
-      - By default, only the first name that back the I(hostnames) list is returned.
+      - By default, the first name that matches an entry of the I(hostnames) list is returned.
       - Turn this flag on if you don't mind having duplicated entries in the inventory
         and you want to get all the hostnames that match.
     type: bool
--- a/plugins/inventory/aws_ec2.py
+++ b/plugins/inventory/aws_ec2.py
@@ -64,6 +64,13 @@ options:
         type: str
         default: '_'
         required: False
+  allow_duplicated_hosts:
+    description:
+      - By default, only the first name that back the I(hostnames) list is returned.
+      - Turn this flag on if you don't mind having duplicated entries in the inventory
+        and you want to get all the hostnames that match.
+    type: bool
+    default: False
   filters:
     description:
       - A dictionary of filter value pairs.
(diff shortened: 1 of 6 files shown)


plugins/inventory/aws_ec2.py
```diff
@@ -64,6 +64,13 @@
         type: str
         default: '_'
         required: False
+  allow_duplicated_hosts:
+    description:
+      - By default, only the first name that back the I(hostnames) list is returned.
+      - Turn this flag on if you don't mind having duplicated entries in the inventory
+        and you want to get all the hostnames that match.
+    type: bool
+    default: False
```
alinabuzachis, line 73: Please add `version_added: 5.0.0`

alinabuzachis (commented):
@goneri Could you also please update the documentation with this option https://github.com/ansible-collections/amazon.aws/blob/main/docs/docsite/rst/aws_ec2_guide.rst ?

goneri:
recheck

goneri:
recheck

alinabuzachis:
recheck

plugins/inventory/aws_ec2.py
```diff
@@ -64,6 +64,14 @@
         type: str
         default: '_'
         required: False
+  allow_duplicated_hosts:
+    description:
+      - By default, only the first name that back the I(hostnames) list is returned.
```
jillr, line 69:
```suggestion
      - By default, only the first name that appears in the I(hostnames) list is returned.
```
Do I understand your meaning here?

plugins/inventory/aws_ec2.py
```diff
@@ -176,6 +184,9 @@
     separator: '-'       # Hostname will be aws-test_literal
     prefix: 'aws'
 
+# Returns all the hostnames for a given instance
+allow_duplicated_hosts: False
```
jillr, line 188:
```suggestion
allow_duplicated_hosts: True
```

plugins/inventory/aws_ec2.py
goneri: I should have said "matches".

By default, the first name that matches  one of the entry of the I(hostnames) list is returned.

plugins/inventory/aws_ec2.py
goneri: My understanding was the the new behaviour is a bug and we would like to switch back to what we've had in 4.0.0.

alinabuzachis: The fix in this PR https://github.com/ansible-collections/amazon.aws/pull/862 should do this (that is how it should be, it's not a bug really).

1. Fix returned hosts when
```
hostnames:
- tag:Tag1,Tag2
```
2. Fix returned hosts when
```
hostnames:
- tag:Tag1
- tag:Tag2
```
3. Fix returned hosts when
```
hostnames:
- tag:Tag1=Test1,Tag2=Test2
```

Given and EC2 instance with the following tags
```
tags:
  Tag1: Test1
  Tag2: Test2
```
Instead of only returning
```
"aws_ec2": {
    "hosts": [
        "Test1"
    ]
},
"tag_Name_instance_02": {
    "hosts": [
        "Test1"
    ]
},
"tag_Tag1_Test1": {
    "hosts": [
        "Test1"
    ]
},
"tag_Tag2_Test2": {
    "hosts": [
        "Test1"
    ]
}
```
It returns now

```
{
"_meta": {
   "hostvars": {
         "Test1": {
                "ami_launch_index": 0,
                "ansible_host": "10.210.0.101",
                   ...},
         "Test2": {
                 "ami_launch_index": 0,
                 "ansible_host": "10.210.0.101",
                   ...},
    },
    "all": {
        "children": [
            "aws_ec2",
            "tag_Name_instance_02",
            "tag_Tag1_Test1",
            "tag_Tag2_Test2",
            "ungrouped"
        ]
    },
    "aws_ec2": {
        "hosts": [
            "Test1",
            "Test2"
        ]
    },
    "tag_Name_instance_02": {
        "hosts": [
            "Test1",
            "Test2"
        ]
    },
    "tag_Tag1_Test1": {
        "hosts": [
            "Test1",
            "Test2"
        ]
    },
    "tag_Tag2_Test2": {
        "hosts": [
            "Test1",
            "Test2"
        ]
    }
}

The problem is it altered the way the hosts were returned introducing a breaking change for the users unfortunately. So, for some users this behaviour is useful while for some others it is not. What about @tremble's suggestion here https://github.com/ansible-collections/amazon.aws/pull/862#issuecomment-1233997139 ?

Comment (goneri) [reply plugins/inventory/aws_ec2.py:188]: This PR implements what @markuman describe here https://github.com/ansible-collections/amazon.aws/pull/862#issuecomment-1233758859

File: plugins/inventory/aws_ec2.py [hunk 4]
```diff
@@ -689,20 +732,29 @@ def _query(self, regions, include_filters, exclude_filters, strict_permissions):
...
+                hostnames=hostnames,
+                allow_duplicated_hosts=allow_duplicated_hosts)
             self.inventory.add_child('all', group)
 
-    def _add_hosts(self, hosts, group, hostnames):
+    def _add_hosts(self, hosts, group, hostnames, allow_duplicated_hosts=False):
         '''
             :param hosts: a list of hosts to be added to a group
             :param group: the name of the group to which the hosts belong
             :param hostnames: a list of hostname destination variables in order of preference
```
alinabuzachis, line 749: Could you please update this with allow_duplicated_hosts=False?

alinabuzachis (approved):
Left one comment, otherwise, lgtm! Thank you @goneri

plugins/inventory/aws_ec2.py
markuman: does anything speak against releasing it with a 4er version?

markuman (approved):

plugins/inventory/aws_ec2.py
jillr: I might just be confused. I would expect `allow_duplicated_hosts: False` to only return unique names.  And I would expect what markuman suggested (`unique_hostnames: true`) to do the same thing.

In this example the comment reads:
`# Returns all the hostnames for a given instance` 
But the example parameter is `allow_duplicated_hosts: False` - which I think would not return "all" hostnames, because it would only return unique / non-duplicate names.

Is my logic wrong?

jillr (commented):
The code LGTM, just a little confused on the naming and default value. Thanks @goneri!

PR #1068, opened 2022-09-26
ignore-2.15: ignore route53's get state

goneri:
Depends-On: https://github.com/ansible-collections/amazon.aws/pull/1066

The error is:

```
ERROR: plugins/modules/route53.py:0:0: parameter-state-invalid-choice: Argument 'state' includes the value 'get' as a choice
```

### DIFF
--- a/tests/sanity/ignore-2.15.txt
+++ b/tests/sanity/ignore-2.15.txt
@@ -1,0 +1,1 @@
+plugins/modules/route53.py validate-modules:parameter-state-invalid-choice # route53_info needs improvements before we can deprecate this


felixfontein (approved):

goneri:
recheck

felixfontein:
Hmm, some of the modules depend on community.aws module_utils, which is fixed by #1066.

felixfontein:
I just noticed there's also #1063 for a similar change.

goneri:
Thank you @felixfontein

goneri:
See @felixfontein comment.

PR #1133, opened 2022-10-06
Add metrics and extended_statistic keys to cloudwatch module

GomathiselviS:

To support https://issues.redhat.com/browse/ACA-638 , a new key metric ( a list of dicts) is added to the cloudwatch module

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

cloudwatch.py

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/cloudwatch_metric_alarm.py
+++ b/plugins/modules/cloudwatch_metric_alarm.py
@@ -42,12 +42,76 @@ options:
           - Unique name for the alarm.
         required: true
         type: str
-    metric:
+    metric_name:
         description:
           - Name of the monitored metric (e.g. C(CPUUtilization)).
           - Metric must already exist.
         required: false
         type: str
+        aliases: ['metric']
+    metrics:
+        description:
+          - An array of MetricDataQuery structures that enable 
+            you to create an alarm based on the result of a metric math expression.
+        type: list
+        elements: dict
+        suboptions:
+            id:
+                description:
+                  - A short name used to tie this object to the results in the response. 
+                type: str
+            metricstat:
+                description: The metric to be returned, along with statistics, period, and units. 
+                type: dict
+                suboptions:
+                    namespace:
+                        description: The namespace of the metric.
+                        type: str
+                    metricname:
+                        description: The name of the metric. 
+                        type: str
+                        required: True
+                    dimensions:
+                        description: a name/value pair that is part of the identity of a metric.
+                        type: list
+                        elements: dict
+                        suboptions:
+                            name:
+                                description: The name of the dimension.
+                                type: str
+                                required: True
+                            value:
+                                description: The value of the dimension.
+                                type: str
+                                required: True
+                    period:
+                        description: The granularity, in seconds, of the returned data points. 
+                        type: int
+                        required: True
+                    stat:
+                        description: The statistic to return. It can include any CloudWatch statistic or extended statistic.
+                        type: str
+                        required: True
+                    unit:
+                        description: Unit to use when storing the metric.
+                        type: str
+              expression:
+                  description: 
+                    - This field can contain either a Metrics Insights query,
+                      or a metric math expression to be performed on the returned data.
+                  type: str
+              label:
+                  description: A human-readable label for this metric or expression. 
+                  type: str
+              return_data:
+                  description: This option indicates whether to return the timestamps and raw data values of this metric.
+                  type: bool
+              period:
+                  description: The granularity, in seconds, of the returned data points. 
+                  type: int
+              account_id:
+                  description: The ID of the account where the metrics are located, if this is a cross-account alarm.
+                  type: str
     namespace:
         description:
           - Name of the appropriate namespace (C(AWS/EC2), C(System/Linux), etc.), which determines the category it will appear under in CloudWatch.
@@ -60,6 +124,9 @@ options:
         required: false
         choices: ['SampleCount','Average','Sum','Minimum','Maximum']
         type: str
+    extended_statistic:
+        description: The percentile statistic for the metric specified in the metric name.
+        type: string
     comparison:
         description:
           - Determines how the threshold value is compared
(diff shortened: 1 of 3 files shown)


plugins/modules/cloudwatch_metric_alarm.py
```diff
@@ -59,6 +139,10 @@
         required: false
         choices: ['SampleCount','Average','Sum','Minimum','Maximum']
         type: str
+    extended_statistic:
+        description: The percentile statistic for the metric specified in the metric name.
+        type: str
+        required: false
```
abikouo, line 145:
```suggestion
        required: false
        version_added: "5.1.0"
```

plugins/modules/cloudwatch_metric_alarm.py
```diff
@@ -41,12 +41,92 @@
...
           - Metric must already exist.
         required: false
         type: str
+        aliases: ['metric']
+    metrics:
+        description:
+          - An array of MetricDataQuery structures that enable
+            you to create an alarm based on the result of a metric math expression.
+        type: list
+        required: false
```
abikouo, line 56:
```suggestion
        required: false
        version_added: "5.1.0"
```

abikouo (changes requested):

plugins/modules/cloudwatch_metric_alarm.py
```diff
@@ -340,6 +471,16 @@ def main():
     params['InsufficientDataActions'] = module.params.get('insufficient_data_actions', [])
     params['OKActions'] = module.params.get('ok_actions', [])
     params['TreatMissingData'] = module.params.get('treat_missing_data')
+    if module.params.get('metrics'):
+        params['Metrics'] = []
```
goneri, line 475: I believe you can just do:
```python
params['Metrics'] = snake_dict_to_camel_dict(module.params.get('metrics'), capitalize_first=True)
```


```
>>> snake_dict_to_camel_dict([{"snaky_the_snack": "a", "SophieTheSnicker": "b"}], 1)
[{'SnakyTheSnack': 'a', 'Sophiethesnicker': 'b'}]
```

plugins/modules/cloudwatch_metric_alarm.py
```diff
@@ -204,18 +310,19 @@
...
-        params['Dimensions'] = fixed_dimensions
+    if params.get('Dimensions'):
+        if not isinstance(params['Dimensions'], list):
+            fixed_dimensions = []
+            for key, value in params['Dimensions'].items():
+                fixed_dimensions.append({'Name': key, 'Value': value})
+            params['Dimensions'] = fixed_dimensions
 
     if not alarms['MetricAlarms']:
         try:
```
gravesm, line 328: I know this wasn't part of this PR, but shouldn't the `changed = True` be outside this try/except block?

plugins/modules/cloudwatch_metric_alarm.py
```diff
@@ -340,6 +471,14 @@ def main():
     params['InsufficientDataActions'] = module.params.get('insufficient_data_actions', [])
     params['OKActions'] = module.params.get('ok_actions', [])
     params['TreatMissingData'] = module.params.get('treat_missing_data')
+    if module.params.get('metrics'):
+        params['Metrics'] = snake_dict_to_camel_dict(module.params['metrics'], capitalize_first=True)
+    if module.params.get('extended_statistic'):
+        params['ExtendedStatistic'] = module.params.get('extended_statistic')
+
+    for key, value in list(params.items()):
```
gravesm, line 479: Why use `list()` here?

GomathiselviS: list() is used here because it preserves the keys & values as they appeared when the list was created. If not, the usage of `del` inside the loop will result in `RuntimeError: dictionary changed size during iteration`

plugins/modules/cloudwatch_metric_alarm.py
GomathiselviS: yes. Only the API call should be inside the try/except block. Do you think we should make this fix in a separate PR, so that there will be clarity on what fixes are going in via a PR?

gravesm: Actually, this is fine. I think I read this the wrong way the first time and thought `changed` wouldn't be set to `True` in check mode. You can disregard.

plugins/modules/cloudwatch_metric_alarm.py
gravesm: Ah, right. You could use a dictionary comprehension here which would be a little more pythonic, but either way is fine.

plugins/modules/cloudwatch_metric_alarm.py
```diff
@@ -310,22 +429,39 @@ def main():
...
+        ['metric_name', 'metrics'],
+        ['dimensions', 'metrics'],
+        ['period', 'metrics'],
+        ['namespace', 'metrics'],
+        ['statistic', 'metrics'],
+        ['extended_statistic', 'metrics'],
+        ['unit', 'metrics'],
+        ['statistic', 'extended_statistic'],
+    ]
+    #mutually_exclusive = []
```
alinabuzachis: I guess you can delete this commented line.

plugins/modules/cloudwatch_metric_alarm.py
alinabuzachis: Why are you deleting None values?

alinabuzachis: Why are you deleting None values? I guess it would be easier to set a params only if defined instead. Otherwise, you have to cover that functionality with a unit test.

plugins/modules/cloudwatch_metric_alarm.py
```diff
@@ -310,22 +424,39 @@ def main():
                                        'Terabytes/Second', 'Bits/Second', 'Kilobits/Second', 'Megabits/Second', 'Gigabits/Second',
                                        'Terabits/Second', 'Count/Second', 'None']),
         evaluation_periods=dict(type='int'),
+        extended_statistic=dict(type='str'),
```
alinabuzachis, line 427: It should be an integer.

plugins/modules/cloudwatch_metric_alarm.py
GomathiselviS: @alinabuzachis The keys that are not defined in the task will be set to 'None' value. I am deleting the None values in order to satisfy the mutually exclusive condition. I will add a test case to test the mutually exclusive condition to the integration test suite, for now.

abikouo (approved):

PR #1140, opened 2022-10-07
ec2_metadata_facts: avoid an AttributeError

goneri:
`ami-launch-index` return an `int`. `json.load()` returns the same
value and the same type. This was causing an `AttributeError` exception
later in the code.

We now properly address the exception.

The commit also re-enables the functional test.


goneri:
See: https://github.com/ansible-collections/amazon.aws/pull/1138

PR #1186, opened 2022-10-20
ec2_metadata_facts: Add support to query instance tags in metadata

### LINKED ISSUE #684: Implement instance metadata tags support
### Summary

I would to enable instance metadata tags using `ec2_instance`. Additionally I'd like to be able to query those tags using `ec2_metadata_facts`

### Issue Type

Feature Idea

### Component Name

ec2_instance, ec2_metadata_facts

### Additional Information

This feature will help me to **_enable_** the ability to query instance metadata tags, as well as actually **_query_** instance metadata tags. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#work-with-tags-in-IMDS

### Code of Conduct

mandar242:
##### SUMMARY

Fixes #684 

Added support to be able to query instance tags using ec2_metadata_facts.

This PR adds a field in returned `ansible_facts` named  ansible_ec2_instance_tags_keys.

Sample
```
"ansible_ec2_instance_tags_keys": [
        "Name",
        "snake_case_key"
],
```

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ec2_metadata_facts

##### ADDITIONAL INFORMATION

Support to enable instance metadata tags already exists in [amazon.aws.ec2_instance](https://github.com/ansible-collections/amazon.aws/blob/main/plugins/modules/ec2_instance.py#L346-L353)

### DIFF
--- a/plugins/modules/ec2_metadata_facts.py
+++ b/plugins/modules/ec2_metadata_facts.py
@@ -259,6 +259,7 @@ ansible_facts:
             type: list
             elements: str
             sample: ["tagKey1", "tag_key2"]
+            version_added: 5.1.0
         ansible_ec2_instance_type:
             description: The type of the instance.
             type: str
--- a/plugins/modules/ec2_metadata_facts.py
+++ b/plugins/modules/ec2_metadata_facts.py
@@ -252,6 +252,13 @@ ansible_facts:
             description: The purchasing option of the instance.
             type: str
             sample: "on-demand"
+        ansible_ec2_instance_tags_keys:
+            description:
+                - The list of tags keys of the instance.
+                - Returns empty list if access to tags (InstanceMetadataTags) in instance metadata is not enabled.
+            type: list
+            elements: str
+            sample: ["tagKey1", "tag_key2"]
         ansible_ec2_instance_type:
             description: The type of the instance.
             type: str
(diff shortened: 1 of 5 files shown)


gravesm (commented):
This also needs a changelog fragment.

plugins/modules/ec2_metadata_facts.py
```diff
@@ -576,6 +594,10 @@ def run(self):
         data.update(dyndata)
         data = self.fix_invalid_varnames(data)
 
+        instance_tags_keys = self._fetch(self.uri_instance_tags)
```
gravesm, line 597: My understanding is that tags are not enabled on instance metadata by default. Isn't this going to fail in that case?

goneri: For the record, this is what I get with an instance and 2 tags:

```
        "ansible_ec2_tags_instance_Name": "test-goneri-tag",
        "ansible_ec2_tags_instance_foo": "var",
```

tests/integration/targets/ec2_metadata_facts/playbooks/setup.yml
```diff
@@ -115,8 +115,11 @@
       network:
         assign_public_ip: true
         delete_on_termination: true
+      metadata_options:
```
goneri, line 118: The instance type is wrong since you need Nitro support (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances). I personally don't care if there is functional test or not since the code follow the same logic than the rest of the facts (assuming here you just use `self.fetch()`).

plugins/modules/ec2_metadata_facts.py
mandar242: @gravesm , it doesn't fail, it returns `str` "None" but I wasn't sure how to handle it

in case if the instance metadata tags are enabled, we are returning a `list` of tag keys like `"ansible_ec2_instance_tags_keys": [ "Name", "snake_case_key" ]` , so I decided to return `None` in a list like `"ansible_ec2_instance_tags_keys": [ "None" ]`, to maintain the output data type. 

should we want to return `ansible_ec2_instance_tags_keys: "None"` instead?

mandar242: > For the record, this is what I get with an instance and 2 tags:
>         "ansible_ec2_tags_instance_Name": "test-goneri-tag",
>         "ansible_ec2_tags_instance_foo": "var",

@goneri , these keys were being returned before the code change in this PR, but there was no single output param to access all the tag keys earlier
with the code change in this PR, we return new output param `ansible_ec2_instance_tags_keys` which is a list of str having tag keys

gravesm: > should we want to return ansible_ec2_instance_tags_keys: "None" instead?

I think we should return `ansible_ec2_instance_tags_keys: []`, which is what the documentation for the return value suggests it will do already.

mandar242: Fixed @gravesm

goneri: Thank you @mandar242 for the clarification, I was far away.

mandar242:
recheck

plugins/modules/ec2_metadata_facts.py
```diff
@@ -252,6 +252,13 @@
             description: The purchasing option of the instance.
             type: str
             sample: "on-demand"
+        ansible_ec2_instance_tags_keys:
+            description:
+                - The list of tags keys of the instance.
+                - Returns empty list if access to tags (InstanceMetadataTags) in instance metadata is not enabled.
```
alinabuzachis, line 258: Please add `version_added: 5.1.0`

mandar242: Added.

mandar242:
recheck

gravesm (approved):

goneri (approved):

alinabuzachis (approved):

PR #1252, opened 2022-11-08
ec2_ami_info/tests: add unit-tests

mandar242:
##### SUMMARY

- Add the unit-test coverage of the ec2_ami_info module.
- break up `list_ec2_images` to move `ec2_client.describe_image_attribute` and `ec2_client.describe_images` to separate methods.
- Add separate method to build request arguments dict
- Do not pass module when we can use an exception to handle the API failures
- Refactor ec2_ami_info module

##### COMPONENT NAME

ec2_ami_info

### DIFF
--- a/plugins/modules/ec2_ami_info.py
+++ b/plugins/modules/ec2_ami_info.py
@@ -217,13 +217,6 @@ from ansible_collections.amazon.aws.plugins.module_utils.ec2 import boto3_tag_li
 
 def build_request_args(executable_users, filters, image_ids, owners):
 
-    request_args = {
-        'ExecutableUsers': [str(user) for user in executable_users],
-        'Filters': ansible_dict_to_boto3_filter_list(filters),
-        'ImageIds': [str(image_id) for image_id in image_ids],
-        'Owners': [str(owner) for owner in owners],
-    }
-
     # describe_images is *very* slow if you pass the `Owners`
     # param (unless it's self), for some reason.
     # Converting the owners to filters and removing from the
@@ -241,6 +234,12 @@ def build_request_args(executable_users, filters, image_ids, owners):
             if 'owner-alias' not in filters:
                 filters['owner-alias'] = list()
             filters['owner-alias'].append(owner)
+
+    request_args = {
+        'ExecutableUsers': [str(user) for user in executable_users],
+        'Filters': ansible_dict_to_boto3_filter_list(filters),
+        'ImageIds': [str(image_id) for image_id in image_ids],
+    }
 
     request_args = {k: v for k, v in request_args.items() if v}
 
--- a/plugins/modules/ec2_ami_info.py
+++ b/plugins/modules/ec2_ami_info.py
@@ -215,13 +215,14 @@ from ansible_collections.amazon.aws.plugins.module_utils.ec2 import ansible_dict
 from ansible_collections.amazon.aws.plugins.module_utils.ec2 import boto3_tag_list_to_ansible_dict
 
 
-def list_ec2_images(ec2_client, module):
+def build_request_args(executable_users, filters, image_ids, owners):
 
-    image_ids = module.params.get("image_ids")
-    owners = module.params.get("owners")
-    executable_users = module.params.get("executable_users")
-    filters = module.params.get("filters")
-    owner_param = []
+    request_args = {
+        'ExecutableUsers': [str(user) for user in executable_users],
+        'Filters': ansible_dict_to_boto3_filter_list(filters),
+        'ImageIds': [str(image_id) for image_id in image_ids],
+        'Owners': [str(owner) for owner in owners],
+    }
 
     # describe_images is *very* slow if you pass the `Owners`
     # param (unless it's self), for some reason.
@@ -235,17 +236,20 @@ def list_ec2_images(ec2_client, module):
             filters['owner-id'].append(owner)
         elif owner == 'self':
             # self not a valid owner-alias filter (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html)
-            owner_param.append(owner)
+            request_args['Owners'].append(owner)
         else:
             if 'owner-alias' not in filters:
                 filters['owner-alias'] = list()
             filters['owner-alias'].append(owner)
 
-    filters = ansible_dict_to_boto3_filter_list(filters)
+    request_args = {k: v for k, v in request_args.items() if v}
 
+    return request_args
+
+
+def list_ec2_images(ec2_client, module, request_args):
     try:
-        images = ec2_client.describe_images(aws_retry=True, ImageIds=image_ids, Filters=filters, Owners=owner_param,
-                                            ExecutableUsers=executable_users)
+        images = ec2_client.describe_images(aws_retry=True, **request_args)
         images = [camel_dict_to_snake_dict(image) for image in images["Images"]]
     except (ClientError, BotoCoreError) as err:
         module.fail_json_aws(err, msg="error describing images")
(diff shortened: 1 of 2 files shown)


mandar242:
recheck

plugins/modules/ec2_ami_info.py
```diff
@@ -235,52 +241,82 @@ def list_ec2_images(ec2_client, module):
...
+
+    return request_args
+
+
+def get_images(ec2_client, request_args):
+    try:
+        images = ec2_client.describe_images(aws_retry=True, **request_args)
+    except (ClientError, BotoCoreError) as err:
+        raise AmiInfoFailure(err, "error describing images")
+    return images
```
abikouo, line 262:
```suggestion
        return ec2_client.describe_images(aws_retry=True, **request_args)
    except (ClientError, BotoCoreError) as err:
        raise AmiInfoFailure(err, "error describing images")
```

plugins/modules/ec2_ami_info.py
```diff
@@ -235,52 +241,82 @@ def list_ec2_images(ec2_client, module):
...
+    images = get_images(ec2_client, module, request_args)["Images"]
+    images = [camel_dict_to_snake_dict(image) for image in images]
+
     for image in images:
         try:
             image['tags'] = boto3_tag_list_to_ansible_dict(image.get('tags', []))
             if module.params.get("describe_image_attributes"):
-                launch_permissions = ec2_client.describe_image_attribute(aws_retry=True, Attribute='launchPermission',
-                                                                         ImageId=image['image_id'])['LaunchPermissions']
+                launch_permissions = get_image_attribute(ec2_client, image)['LaunchPermissions']
```
abikouo, line 283:
```suggestion
                launch_permissions = get_image_attribute(ec2_client, image).get('LaunchPermissions')
```

tests/unit/plugins/modules/test_ec2_ami_info.py
```diff
@@ -0,0 +1,253 @@
...
+from unittest.mock import MagicMock, patch, ANY, call
+import botocore.exceptions
+import pytest
+
+from ansible_collections.amazon.aws.plugins.modules import ec2_ami_info
+
+module_name = "ansible_collections.amazon.aws.plugins.modules.ec2_ami_info"
+
+
+@pytest.mark.parametrize("executable_users,filters,image_ids,owners,expected", [
```
abikouo, line 16: `executable_users` has only one value, this can be removed from the parameterized or add other values to increase coverage

tests/unit/plugins/modules/test_ec2_ami_info.py
```diff
@@ -0,0 +1,253 @@
...
+        ],
+    }
+
+    request_args = {'ImageIds': ['ami-1234567890']}
+
+    get_images_result = ec2_ami_info.get_images(ec2_client, module, request_args)
+
+    ec2_client.describe_images.call_count == 2
+    ec2_client.describe_images.assert_called_with(aws_retry=True, **request_args)
+    assert len(get_images_result['Images']) == 1
```
abikouo, line 65:
```suggestion
    assert get_images_result == ec2_client.describe_images.return_value
```

tests/unit/plugins/modules/test_ec2_ami_info.py
```diff
@@ -0,0 +1,253 @@
...
+    ([], {'owner-alias': 'test_ami_owner'}, [], ['1234567890'], {'Filters': [{'Name': 'owner-alias',
+     'Values': ['test_ami_owner']}, {'Name': 'owner-id', 'Values': ['1234567890']}]}),
+    ([], {'is-public': True}, [], [], {'Filters': [{'Name': 'is-public', 'Values': ['true']}]}),
+    ([], {}, [], ['self'], {'Owners': ['self']})])
+def test_build_request_args(executable_users, filters, image_ids, owners, expected):
+    assert ec2_ami_info.build_request_args(
+        executable_users, filters, image_ids, owners) == expected
+
+
+def test_get_images():
```
abikouo, line 29: you should add a test case for the failure

tests/unit/plugins/modules/test_ec2_ami_info.py
```diff
@@ -0,0 +1,253 @@
...
+    request_args = {'ImageIds': ['ami-1234567890']}
+
+    get_images_result = ec2_ami_info.get_images(ec2_client, module, request_args)
+
+    ec2_client.describe_images.call_count == 2
+    ec2_client.describe_images.assert_called_with(aws_retry=True, **request_args)
+    assert len(get_images_result['Images']) == 1
+
+
+def test_get_image_attribute():
```
abikouo, line 68: same here for the failure test case

tests/unit/plugins/modules/test_ec2_ami_info.py
```diff
@@ -0,0 +1,253 @@
...
+        "name": "test-ami-uefi-boot",
+        "owner_id": "1234567890",
+        "platform_details": "Linux/UNIX"
+    }
+
+    get_image_attribute_result = ec2_ami_info.get_image_attribute(ec2_client, image)
+
+    ec2_client.describe_image_attribute.call_count == 1
+    ec2_client.describe_image_attribute.assert_called_with(aws_retry=True, Attribute='launchPermission', ImageId=image['image_id'])
+    assert len(get_image_attribute_result['LaunchPermissions']) == 2
```
abikouo, line 109: you should test as much as possible the full result

tests/unit/plugins/modules/test_ec2_ami_info.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+@patch(module_name + '.get_image_attribute')
+@patch(module_name + '.get_images')
+def test_list_ec2_images(m_get_images, m_get_image_attribute):
+    ec2_client = MagicMock()
```
abikouo, line 115: as this is repeated many times, you can create a fixture

abikouo (dismissed):
Good job overall, just a few remarks.
Thanks Mandar

tests/unit/plugins/modules/test_ec2_ami_info.py
goneri: This structure was passed to Mock and here we've got what Mock has returned. If something goes wrong, it will be within Mock which is not what we're testing here. I don't think this will help.

plugins/modules/ec2_ami_info.py
goneri: In this case you need to default on `[]` to avoid a `TypeError` later:

```
launch_permissions = get_image_attribute(ec2_client, image).get('LaunchPermissions', [])
```

This being said, the use of .get() may also be a bit risky and hide a real problem, like e.g if the API call return something totally unexpected.

tests/unit/plugins/modules/test_ec2_ami_info.py
abikouo: this is not true, the mock is just inside the function we are testing, so if inside the testing function the result returned by the mock function is modified this test will not failed but it should

tests/unit/plugins/modules/test_ec2_ami_info.py
mandar242: yes, already added https://github.com/ansible-collections/amazon.aws/pull/1252/files#diff-6192636df21722bd5081088ae5e21132bb06367a3b687c2ea8e64553759a19acR237-R253

tests/unit/plugins/modules/test_ec2_ami_info.py
mandar242: It is used only in `test_list_ec2_images`, I feel fixture could be unnecessary.

mandar242: Added.

tests/unit/plugins/modules/test_ec2_ami_info.py
mandar242: Added a assertion with `executable_users`.

tests/unit/plugins/modules/test_ec2_ami_info.py
mandar242: Fixed.

tests/unit/plugins/modules/test_ec2_ami_info.py
mandar242: yes, added.

goneri (approved):

goneri:
regate

PR #1270, opened 2022-11-16
elbv2: respect UseExistingClientSecret

markuman:
### SUMMARY

Since amazon.aws 5.0.0, `elb_application_lb` runs into an exception, when using `Type: authenticate-oidc` in a rule, even when `UseExistingClientSecret: True` parameter is given. That works as expected with amazon.aws 4.x.x.

The logic gets broken in  https://github.com/ansible-collections/amazon.aws/pull/940  

Basically AWS won't return both, `UseExistingClientSecret` and  `ClientSecret`.   
But when requesting against boto3,  both parameters are mutually exclusive!  

When the user set `UseExistingClientSecret: True`, the `ClientSecret` must be removed for the request.  
When the user does not set `UseExistingClientSecret` or set it to `False`,  the `UseExistingClientSecret` must be included in the request.

While diving deeper, I've noticed a basic change detection problem for default values, that are not requested, but AWS will return them. I've summerized it in #1284  
However, this PR does not target #1284, it just fixes the exception and restores the functionality and hotfix the change-detection only for `Type: authenticate-oidc`.

##### origin PR description
---

```
The error was: botocore.errorfactory.InvalidLoadBalancerActionException: An error occurred (InvalidLoadBalancerAction) when calling the ModifyRule operation: You must either specify a client secret or set UseExistingClientSecret to true
```
`UseExistingClientSecret` is not respected anymore since a.a 5

Introduced in https://github.com/ansible-collections/amazon.aws/pull/940

Furthermore, AWS returns also `Scope` and ` SessionTimeout` parameters that are filled with default values if not requested.

```
'Scope': 'openid',
'SessionTimeout': 604800,
```

That make the module always returns a `change`, if they are not requested.  
This fix does not break backwards compatibility, because the values are already set by aws, when not requested yet.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/elbv2.yml

##### ADDITIONAL INFORMATION

```yml
          - Conditions:
              - Field: host-header
                Values:
                  - some.tld
              - Field: path-pattern
                Values:
                  - "/admin/*"
            Actions:
              - Type: authenticate-oidc
                Order: 1
                AuthenticateOidcConfig:
                  Issuer: https://login.microsoftonline.com/32rw-ewad53te-ef/v2.0
                  AuthorizationEndpoint: https://login.microsoftonline.com/324re-dafs6-6tw/oauth2/v2.0/authorize
                  TokenEndpoint: https://login.microsoftonline.com/432535ez-rfes-32543ter/oauth2/v2.0/token
                  UserInfoEndpoint: https://graph.microsoft.com/oidc/userinfo
                  ClientId: fasgd-235463-fsgd-243
                  ClientSecret: "{{ lookup('onepassword', 'some cool secret', vault='some important vault') }}"
                  SessionCookieName: AWSELBAuthSessionCookie
                  OnUnauthenticatedRequest: authenticate
                  UseExistingClientSecret: True
              - TargetGroupName: "{{ some_tg }}"
                Type: forward
                Order: 2
```

```paste below

```

### DIFF
--- a/changelogs/fragments/1270-elbv2-fixes.yml
+++ b/changelogs/fragments/1270-elbv2-fixes.yml
@@ -1,0 +1,3 @@
+bugfixes:
+  - module_utils/elbv2 - respect ``UseExistingClientSecret`` parameter in ``authenticate-oidc`` rules (https://github.com/ansible-collections/amazon.aws/pull/1270).
+  - module_utils/elbv2 - fix change detection by adding default values for ``Scope`` and ``SessionTimeout`` parameters in ``authenticate-oidc`` rules (https://github.com/ansible-collections/amazon.aws/pull/1270).
--- a/plugins/module_utils/elbv2.py
+++ b/plugins/module_utils/elbv2.py
@@ -89,17 +89,31 @@ def _prune_ForwardConfig(action):
     return newAction
 
 
-# the AWS api won't return the client secret, so we'll have to remove it
-# or the module will always see the new and current actions as different
-# and try to apply the same config
+# remove the client secret if UseExistingClientSecret, because aws won't return it
+# add default values when they are not requested
 def _prune_secret(action):
     if action['Type'] != 'authenticate-oidc':
         return action
 
+    if not action['AuthenticateOidcConfig'].get('Scope', False):
+        action['AuthenticateOidcConfig']['Scope'] = 'openid'
+
+    if not action['AuthenticateOidcConfig'].get('SessionTimeout', False):
+        action['AuthenticateOidcConfig']['SessionTimeout'] = 604800
+
     if action['AuthenticateOidcConfig'].get('UseExistingClientSecret', False):
         action['AuthenticateOidcConfig'].pop('ClientSecret', None)
-    else:
-        action['AuthenticateOidcConfig'].pop('UseExistingClientSecret')
+
+    return action
+
+
+# while AWS api also won't return UseExistingClientSecret key
+# it must be added, because it's requested and compared
+def _append_use_existing_client_secretn(action):
+    if action['Type'] != 'authenticate-oidc':
+        return action
+
+    action['AuthenticateOidcConfig']['UseExistingClientSecret'] = True
 
     return action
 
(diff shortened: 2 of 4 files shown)


markuman:
Ah okey, the logic is nailed into an unittest...that needs some more work.

markuman:
related to the `changed` detection. this is a minefield when you search for "The default is" on https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html  
there are dozens of possibilities where a rule condition returns default parameters + values, that are not requestes by the user/module.

Screenshot from 2022-11-17 12-25-55

Just `pop()` the values that AWS returns but are not requested may be breaks backwards compatibility. because until the day of now, boto3 is setting those default values if not explicit requested. In other words, when an `elb_application_lb` is deployed via something else (_clicked, terraform, cloudformation, cdk ..._), with different values as the default, our `elb_application_lb` module is currently resetting those value to the defaults when they are not requested. That's the behavior since `elb_application_lb` exists.

Technically the current `rules` behavior must be deprecated and reworked later. parameters that are not requested, must keep the values from what AWS is returning. Or a `purge_rules_parameter_values` parameter must be introduced to take control of what will happen. That will be a larger work item - imo.

However, this is an important PR, because the `elb_application_lb` module fails since amazon.aws 5.0.0, while it was working in `4.3.0` when using a rule action of type `authenticate-oidc`!

cc @tremble @goneri @alinabuzachis @jillr

markuman:
> there are dozens of possibilities where a rule condition returns default parameters + values, that are not requestes by the user/module.

this relates to https://github.com/ansible-collections/community.aws/issues/604

alinabuzachis:
> related to the `changed` detection. this is a minefield when you search for "The default is" on https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html there are dozens of possibilities where a rule condition returns default parameters + values, that are not requestes by the user/module.
> 
> Screenshot from 2022-11-17 12-25-55
> 
> Just `pop()` the values that AWS returns but are not requested may be breaks backwards compatibility. because until the day of now, boto3 is setting those default values if not explicit requested. In other words, when an `elb_application_lb` is deployed via something else (_clicked, terraform, cloudformation, cdk ..._), with different values as the default, our `elb_application_lb` module is currently resetting those value to the defaults when they are not requested. That's the behavior since `elb_application_lb` exists.
> 
> Technically the current `rules` behavior must be deprecated and reworked later. parameters that are not requested, must keep the values from what AWS is returning. Or a `purge_rules_parameter_values` parameter must be introduced to take control of what will happen. That will be a larger work item - imo.
> 
> However, this is an important PR, because the `elb_application_lb` module fails since amazon.aws 5.0.0, while it was working in `4.3.0` when using a rule action of type `authenticate-oidc`!
> 
> cc @tremble @goneri @alinabuzachis @jillr

Can you please open an issue for `rules`  behaviour?

alinabuzachis (approved):

markuman:
> Can you please open an issue for `rules` behaviour?

https://github.com/ansible-collections/amazon.aws/issues/1284

goneri:
Does this PR close #1284? If so can you add `Closes: #1284` in the description.

markuman:
> Does this PR close #1284? If so can you add `Closes: #1284` in the description.

No. It's something what need to be done in the future.  
This PR fixes a botocore exception that was introduced in 5.0.0

markuman:
> Does this PR close #1284? If so can you add `Closes: #1284` in the description.

@goneri I've updates the PR summary. I hope that makes it more clear.

tremble (approved):

PR #1365, opened 2023-02-13
inventory aws_ec2  - add support for AND logic

### LINKED ISSUE #1354: Filters does not support AND logic of multiple tag-key values
### Summary

It seems that there is no way to query ec2 instances via multiple tag-key filters that have AND logic since `filters` is a dict and `include_filters` treats all variations as OR conditions.

### Issue Type

Bug Report

### Component Name

amazon.aws.aws_ec2

### Ansible Version

```console
ansible [core 2.14.2]
```

### Collection Versions

```console
Collection Version
---------- -------
amazon.aws 5.1.0  
```

### AWS SDK versions

```console (paste below)
Name: boto3
Version: 1.26.60

Name: botocore
Version: 1.29.60
```

### Configuration

_No response_

### OS / Environment

_No response_

### Steps to Reproduce

```yaml
plugin: amazon.aws.aws_ec2

hostnames:
  - tag:Name

filters:
  # Applies as OR logic
  # Because filters is a dict can't specify multiple separate tag-key filters
  tag-key:
    - Role
    - Environment
```

```yaml
plugin: amazon.aws.aws_ec2

hostnames:
  - tag:Name

include_filters:
  # Applies as OR logic
  - tag-key: Role
  - tag-key: Environment

```

```yaml
plugin: amazon.aws.aws_ec2

hostnames:
  - tag:Name

include_filters:
  # Also applies as OR logic
  - tag-key: 
      - Role
      - Environment
```

### Expected Results

That there is a way to apply AND logic for multiple tag keys as per the aws command below:

```bash
aws ec2 describe-instances --filter Name=tag-key,Values=Role Name=tag-key,Values=Team
```

This is in line with the AWS documentation which says that multiple filters with the same name are treated with the AND condition:

https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instances.html#options

### Actual Results

Both the `filters` version and `include_filters` return instances that have either the tag Role defined or Team defined, but not explicitly the ones that only have both.

### Code of Conduct

abikouo:
##### SUMMARY

Closes #1354 

##### ISSUE TYPE

- Feature Pull Request


abikouo:
recheck

abikouo:
recheck

plugins/inventory/aws_ec2.py
```diff
@@ -73,8 +73,8 @@
     description:
       - A list of filters. Any instances matching at least one of the filters are included in the result.
       - Available filters are listed here U(http://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instances.html#options).
-      - Every entry in this list triggers a search query. As such, from a performance point of view, it's better to
-        keep the list as short as possible.
+      - Every entry in this list triggers a search query if C(apply_and_logic_on_include_filters) is not set to I(true).
```
alinabuzachis, line 76: I'm not sure though, but shouldn't this be the default behaviour instead of defining a new parameter?
@tremble

abikouo: setting this as default behavior will lead to a breaking change knowing that current behavior is also okay, why then modify it?

tremble (commented):
Apparently I forgot to submit my comment.  I think this PR should be revivable...

PR #1502, opened 2023-05-02
Comment on permission because of botocore.exceptions.ClientError: An error occurred (AccessControlListNotSupported) when calling the PutObject operation: The bucket does not allow ACLs

alinabuzachis:
##### SUMMARY

Comment on permission because of botocore.exceptions.ClientError: An error occurred (AccessControlListNotSupported) when calling the PutObject operation: The bucket does not allow ACLs

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/tests/integration/targets/lambda_layer/tasks/main.yml
+++ b/tests/integration/targets/lambda_layer/tasks/main.yml
@@ -45,7 +45,7 @@
         bucket: "{{ s3_bucket_name }}"
         mode: put
         object: "{{ s3_bucket_object }}"
-        permission: public-read
+        # permission: public-read  # Commented on because botocore.exceptions.ClientError: An error occurred (AccessControlListNotSupported) when calling the PutObject operation: The bucket does not allow ACLs
         src: "{{ zip_file_path }}"
 
     - name: Create lambda layer (check_mode=true)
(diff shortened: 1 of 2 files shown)


alinabuzachis:
I have opened an issue for this https://github.com/ansible-collections/amazon.aws/issues/1503

tests/integration/targets/lambda_layer/tasks/main.yml
```diff
@@ -45,7 +45,7 @@
         bucket: "{{ s3_bucket_name }}"
         mode: put
         object: "{{ s3_bucket_object }}"
-        permission: public-read
+        # permission: public-read  # Commented on because botocore.exceptions.ClientError: An error occurred (AccessControlListNotSupported) when calling the PutObject operation: The bucket does not allow ACLs
```
jillr, line 48: Can you file an issue for us to come back and setup the bucket permissions correctly so that we can test the `permissions` key for objects?  I understand we want to unblock CI for the release in the short term but we should also fix this in the mid term.

alinabuzachis: @jillr  Have filed this issue https://github.com/ansible-collections/amazon.aws/issues/1503

PR #1558, opened 2023-05-19
add wait_for_modify_volume_complete option

mschurenko:
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ec2_vol

##### ADDITIONAL INFORMATION

I ran into the following issue when trying to modify an unattached volume: 

```paste below
boto3_version: 1.26.122
  botocore_version: 1.29.122
  error:
    code: OperationNotPermitted
    message: 'Cannot attach volume vol-02f06b9e2f1e6d1e0 when it is in modification state: MODIFYING'
  msg: 'Error while attaching EBS volume: An error occurred (OperationNotPermitted) when calling the AttachVolume operation: Cannot attach volume vol-02f06b9e2f1e6d1e0 when it is in modification state: MODIFYING'
```
There seems to be no waiter, but if we poll `describe_volumes_modifications` we can wait until `ModificationState` is `complete`.

### DIFF
--- a/plugins/modules/ec2_vol.py
+++ b/plugins/modules/ec2_vol.py
@@ -439,6 +439,22 @@ def update_volume(module, ec2_conn, volume):
             volume["multi_attach_enabled"] = response.get("VolumeModification").get("TargetMultiAttachEnabled")
             volume["throughput"] = response.get("VolumeModification").get("TargetThroughput")
 
+            if module.params.get("wait_for_modify_volume_complete"):
+                mod_state = ""
+                _max_attempts = 100
+                _sleep_secs = 10
+                _attempt = 0
+                while mod_state != "completed":
+                    _attempt += 1
+                    mod_response = ec2_conn.describe_volumes_modifications(VolumeIds=[volume["volume_id"]])
+                    mod_state = mod_response.get("VolumesModifications")[0].get("ModificationState")
+                    # this can take a long time, depending on how much bigger the requested volume size is
+                    if _attempt > _max_attempts:
+                        module.fail_json(
+                            msg=f"Volume {volume['volume_id']} modification state has not reached 'completed' after {_max_attempts * _sleep_secs} seconds."
+                        )
+                    time.sleep(_sleep_secs)
+
     return volume, changed
 
 
@@ -714,6 +730,7 @@ def main():
         outpost_arn=dict(type="str"),
         purge_tags=dict(type="bool", default=True),
         multi_attach=dict(type="bool"),
+        wait_for_modify_volume_complete=dict(default=False, type="bool")
     )
 
     module = AnsibleAWSModule(
@@ -738,6 +755,9 @@ def main():
     volume_type = module.params.get("volume_type")
     throughput = module.params.get("throughput")
     multi_attach = module.params.get("multi_attach")
+    modify_volume = module.params.get("modify_volume")
+    wait_for_modify_volume_complete = module.params.get("wait_for_modify_volume_complete")
+
 
     # Ensure we have the zone or can get the zone
     if instance is None and zone is None and state == "present":
@@ -768,6 +788,9 @@ def main():
 
     if multi_attach is True and volume_type not in ("io1", "io2"):
         module.fail_json(msg="multi_attach is only supported for io1 and io2 volumes.")
+
+    if wait_for_modify_volume_complete is True and modify_volume is False:
+        module.fail_json(msg="wait_for_modify_volume_complete does nothing if modify_volume is False.")
 
     # Set changed flag
     changed = False
--- a/plugins/modules/ec2_vol.py
+++ b/plugins/modules/ec2_vol.py
@@ -108,6 +108,7 @@ options:
     description:
       - Wait for volume modification to complete
     type: bool
+    default: false
     version_added: 6.3.0
 author:
   - "Lester Wade (@lwade)"
--- a/plugins/modules/ec2_vol.py
+++ b/plugins/modules/ec2_vol.py
@@ -455,9 +455,7 @@ def update_volume(module, ec2_conn, volume):
                     mod_state = mod_response.get("VolumesModifications")[0].get("ModificationState")
 
                     if mod_state == "failed":
-                        module.fail_json(
-                            msg=f"Volume {volume['volume_id']} modification has failed."
-                        )
+                        module.fail_json(msg=f"Volume {volume['volume_id']} modification has failed.")
 
                     # this can take a long time, depending on how much bigger the requested volume size is
                     if _attempt > _max_attempts:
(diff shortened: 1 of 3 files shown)


mschurenko:
Hi. Just bumping in case this wasn't noticed.

plugins/modules/ec2_vol.py
```diff
@@ -439,6 +439,22 @@ def update_volume(module, ec2_conn, volume):
             volume["multi_attach_enabled"] = response.get("VolumeModification").get("TargetMultiAttachEnabled")
             volume["throughput"] = response.get("VolumeModification").get("TargetThroughput")
 
+            if module.params.get("wait_for_modify_volume_complete"):
+                mod_state = ""
+                _max_attempts = 100
+                _sleep_secs = 10
+                _attempt = 0
+                while mod_state != "completed":
```
hakbailey, line 447: Would it be worthing adding a check for `failed` state here too and exiting with an error in that case?

hakbailey (changes requested):
Thanks for submitting this PR! Could you add the new option to the [documentation block](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#documentation-block)?

plugins/modules/ec2_vol.py
mschurenko: Hi. I will add checking for `failed`. Great idea. What value should I put for `version_added` in the description?

hakbailey: I think 6.3.0 is our next release.

mschurenko: ok I made those changes. I also increased the polling interval as, even for small volumes, it can take a really long time to finish

hakbailey (approved):

hakbailey (changes requested):
Whoops, just noticed there isn't a changelog fragment for this, could you add that too? Thanks!

mschurenko:
> > Whoops, just noticed there isn't a changelog fragment for this, could you add that too? Thanks!
> 
> So should I add something to `CHANGELOG.rst`?
> 
> Sorry, I see it's documented here https://docs.ansible.com/ansible/latest/community/collection_development_process.html#creating-a-changelog-fragment

Ok I think (hope) I did it correctly.

plugins/modules/ec2_vol.py
```diff
@@ -104,6 +104,12 @@
       - If set, allows to create volume in an Outpost.
     type: str
     version_added: 3.1.0
+  wait_for_modify_volume_complete:
```
alinabuzachis, line 107: I would prefer to add `wait` and `wait_timeout` parameters instead of `wait_for_modify_volume_complete` like most Ansible modules support. For example, https://github.com/ansible-collections/amazon.aws/blob/main/plugins/modules/rds_cluster_snapshot.py#L58.

alinabuzachis (changes requested):

plugins/modules/ec2_vol.py
mschurenko: Ok I made a commit to use `wait` and `wait_timeout` instead

hakbailey:
recheck

plugins/modules/ec2_vol.py
```diff
@@ -439,6 +451,26 @@ def update_volume(module, ec2_conn, volume):
             volume["multi_attach_enabled"] = response.get("VolumeModification").get("TargetMultiAttachEnabled")
             volume["throughput"] = response.get("VolumeModification").get("TargetThroughput")
 
+            if module.params.get("wait"):
+                mod_state = ""
+                _wait_till = module.params.get("wait_timeout") + time.time()
+                while _wait_till > time.time():
+                    mod_response = ec2_conn.describe_volumes_modifications(VolumeIds=[volume["volume_id"]])
```
alinabuzachis, line 458: One small thing. I guess we should enclose  mod_response = ec2_conn.describe_volumes_modifications(VolumeIds=[volume["volume_id"]]) in a try-except block.

plugins/modules/ec2_vol.py
```diff
@@ -769,6 +805,9 @@ def main():
     if multi_attach is True and volume_type not in ("io1", "io2"):
         module.fail_json(msg="multi_attach is only supported for io1 and io2 volumes.")
 
+    if wait is True and modify_volume is False:
+        module.fail_json(msg="wait does nothing if modify_volume is False.")
```
alinabuzachis, line 809: I guess it should work for now. We would probably also like to add the option to wait for deletion in the future.

plugins/modules/ec2_vol.py
mschurenko: Ok, I can add that

plugins/modules/ec2_vol.py
mschurenko: We are using this module currently and have decided to wait for either `completed` or `optimizing` as waiting for `completed` can take a very long time. Should we do the same here? Or should we have the option to wait for one or more modification states?

How about having an option like this?
```yaml
wait_modification_states:
    description:
      - A list of modification states to wait for. Valid options are "completed" and "optimizing"
    type: list
    default: ["completed", "optimizing"]
```

plugins/modules/ec2_vol.py
```diff
@@ -104,6 +104,18 @@
       - If set, allows to create volume in an Outpost.
     type: str
     version_added: 3.1.0
+  wait:
+    description:
+      - Wait for volume modification to complete.
+    type: bool
+    default: false
+    version_added: 6.3.0
```
tremble, line 112:
```suggestion
    version_added: 11.3.0
```

plugins/modules/ec2_vol.py
```diff
@@ -104,6 +104,18 @@
...
+  wait_timeout:
+    description:
+      - How long before wait gives up, in seconds.
+    type: int
+    default: 900
+    version_added: 6.3.0
```
tremble, line 118:
```suggestion
    version_added: 11.3.0
```

plugins/modules/ec2_vol.py
```diff
@@ -104,6 +104,18 @@
       - If set, allows to create volume in an Outpost.
     type: str
     version_added: 3.1.0
+  wait:
+    description:
+      - Wait for volume modification to complete.
```
tremble, line 109:
```suggestion
      - If changes are made, wait for volume modification to complete.
      - Ignored if no changes are made.
```

plugins/modules/ec2_vol.py
tremble: Over in ec2_instance, we did this by adding an additional state.  That gave us the ability to distinguish between "exists but stopped", "exists and has been started" and "exists and reports healthy".  To follow the same pattern, what we could do here is add a "completed" state.  I'd be careful of adding too many extra parameters.

tremble, line 809: This isn't a "failure".  In general, we should be following the rules of idempotency, and re-running the same task with the same parameters should just result in a quick return and "changed=False".

tremble (changes requested):
Sorry, I apparently started a review and forgot to hit submit.

PR #1610, opened 2023-06-12
Backup vault tag bugfix

hakbailey:
##### SUMMARY
Fixes an error raised in `backup_vault` when updating tags on an existing vault by calling the correct boto3 client methods for tagging and untagging a resource. This is not applicable when creating a vault as tags are included in the `create_backup_vault` method params, but when updating a vault we have to call the tag methods individually.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`backup_vault`

##### ADDITIONAL INFORMATION
See the included new integration test task for a command that previously failed and works with the updated code.

### DIFF
--- a/changelogs/fragments/20230612-backup_vault-fix-tag-update.yml
+++ b/changelogs/fragments/20230612-backup_vault-fix-tag-update.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - backup_vault - fix error when updating tags on a backup vault by using the correct boto3 client methods for tagging and untagging backup resources.
+  - backup_vault - fix error when updating tags on a backup vault by using the correct boto3 client methods for tagging and untagging backup resources (https://github.com/ansible-collections/amazon.aws/pull/1610).
--- a/changelogs/fragments/20230612-backup_vault-fix-tag-update.yml
+++ b/changelogs/fragments/20230612-backup_vault-fix-tag-update.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - backup_vault - fix error when updating tags on a backup vault by using the correct boto3 client methods for tagging and untagging backup resources.
--- a/tests/integration/targets/backup_vault/tasks/main.yml
+++ b/tests/integration/targets/backup_vault/tasks/main.yml
@@ -72,20 +72,177 @@
           - backup_vault_result_idem.vault.encryption_key_arn == key.key_arn
           - backup_vault_result_idem.vault.tags.environment == "dev"
 
+    - name: Update AWS Backup Vault - check mode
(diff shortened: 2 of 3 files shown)


changelogs/fragments/20230612-backup_vault-fix-tag-update.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - backup_vault - fix error when updating tags on a backup vault by using the correct boto3 client methods for tagging and untagging backup resources.
```
alinabuzachis, line 3:
```suggestion
  - backup_vault - fix error when updating tags on a backup vault by using the correct boto3 client methods for tagging and untagging backup resources (https://github.com/ansible-collections/amazon.aws/pull/1610).
```

alinabuzachis (commented):
Can you please add some integration tests for this change? Thank you.

hakbailey:
> Can you please add some integration tests for this change? Thank you.

I did, see the new integration task. Is there more I should add?

tests/integration/targets/backup_vault/tasks/main.yml
```diff
@@ -67,7 +70,23 @@
           - backup_vault_result_idem is not changed
           - backup_vault_result_idem.vault.backup_vault_name == backup_vault_name
           - backup_vault_result_idem.vault.encryption_key_arn == key.key_arn
- 
+          - backup_vault_result_idem.vault.tags.environment == "dev"
```
alinabuzachis, line 73: We generally used to add several tests for tags:

- Add tags in check_mode
- Add tags
- Add the same tags (idempotency) in check_mode
- Add the same tags (idempotency)
- Update tags in check_mode
- Update tags 
- Update tags (idempotency) in check_mode
- Update tags (idempotency)
- Similar tests with purge_tags
For example https://github.com/ansible-collections/amazon.aws/blob/main/tests/integration/targets/ec2_vpc_igw/tasks/main.yml

hakbailey: I'm happy to add more test coverage for tagging as part of this PR...however we are also missing both check and idempotency tests for all of the update and delete functionality in this module, and we don't have that level of tagging integration test coverage for any of the other recently added backup modules either. Would it be ok if we create a separate Jira ticket (or tickets) to improve the robustness of tests in these modules in general?

hakbailey: @alinabuzachis I'm so glad you suggested this, adding tests found one more bug! I'll go ahead and create Jira tickets for improving tests on the other modules.

gravesm (approved):

PR #1611, opened 2023-06-13
Backup plan params bugfix

hakbailey:
##### SUMMARY
This updates the `backup_plan` module to remove all `None` values from nested dicts in supplied params. Previously nested `None` values were retained and caused errors when sent through to the boto3 client calls.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`backup_plan`

##### ADDITIONAL INFORMATION
Creating or updating a plan without providing an optional value in a nested dict would result in that value being set to `None`, and it wasn't being removed. For example:
```
    - name: Update Backup plan
      amazon.aws.backup_plan:
        backup_plan_name: my-backup-plan
        rules:
          - rule_name: my-rule
            target_backup_vault_name: my-vault
            schedule_expression: "cron(0 * ? * * *)"
            lifecycle:
              move_to_cold_storage_after_days: 30
```
The optional `lifecycle.delete_after_days` option in the above example would be set to `None` by default, and the removal of `None` values wasn't recursively going through the entire dict. This is now fixed and the above example will work.

### DIFF
--- a/changelogs/fragments/202306012-backup_plan-remove-none-from-nested-params.yml
+++ b/changelogs/fragments/202306012-backup_plan-remove-none-from-nested-params.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - backup_plan - Use existing `scrub_none_values` function from module_utils to remove None values from nested dicts in supplied params. Nested None values were being retained and causing an error when sent through to the boto3 client operation.
--- a/changelogs/fragments/202306012-backup_plan-remove-none-from-nested-params.yml
+++ b/changelogs/fragments/202306012-backup_plan-remove-none-from-nested-params.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - backup_plan - Use existing `scrub_none_values` function from module_utils to remove None values from nested dicts in supplied params. Nested None values were being retained and causing an error when sent through to the boto3 client operation.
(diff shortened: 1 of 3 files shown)


changelogs/fragments/202306012-backup_plan-remove-none-from-nested-params.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - backup_plan - Use existing `scrub_none_values` function from module_utils to remove None values from nested dicts in supplied params. Nested None values were being retained and causing an error when sent through to the boto3 client operation.
```
alinabuzachis, line 3:
```suggestion
  - backup_plan - Use existing `scrub_none_values` function from module_utils to remove None values from nested dicts in supplied params. Nested None values were being retained and causing an error when sent through to the boto3 client operation (https://github.com/ansible-collections/amazon.aws/pull/1611).
```

tests/integration/targets/backup_plan/tasks/main.yml
```diff
@@ -113,13 +113,11 @@
             completion_window_minutes: 150
             lifecycle:
               move_to_cold_storage_after_days: 30
-              delete_after_days: 120
```
alinabuzachis, line 116: Why was this option removed?

alinabuzachis: I would prefer to add an additional test rather than deleting options from an actual one. This way we keep testing all the lifecycle options and when some of them are not set.

alinabuzachis (approved):

abikouo (approved):

PR #1671, opened 2023-07-28
Fix rds_cluster tests on stable-5 branch

gravesm:
##### SUMMARY

This fixes two problems with the `rds_cluster` test target. Ansible 2.16 is removing `ansible.builtin.include` (and has removed it in the current milestone branch), so this has been replaced with `ansible.builtin.include_tasks`. In addition, this updates the engine as Aurora 1 has been retired.

I'm a little confused by the history of this test target. #1493 and #1504 seem to have been merged on the same day, with one overwriting some of the changes of the other. The tests for `rds_cluster` on the `stable-5` branch are currently broken, though, as can be seen in the failure on #1664. If there's a better way to fix this, feel free to suggest something different.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


jillr (approved):
I"m also a little unsure about the PR history, but if I compare these values with the main branch this change seems consistent with how we are testing there.

alinabuzachis (approved):

PR #1672, opened 2023-07-31
[stable-6] Add github workflows (sanity, units, linters and changelog)

alinabuzachis:
##### SUMMARY

Add github workflows to stable-6

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


hakbailey (approved):

.github/workflows/units.yml
```diff
@@ -0,0 +1,68 @@
+---
+name: unit tests
+
+on: [workflow_call]  # allow this workflow to be called from other workflows
+
+jobs:
+  unit-source:
+    uses: ansible-network/github_actions/.github/workflows/unit_source.yml@main
+    with:
+      matrix_exclude: >-
```
gravesm, line 10: Do we actually need this anymore? It seems like we should be able to just use the default excludes.

gravesm: This change won't work because changes in this repo aren't merged with the defaults, they completely overwrite. What I'm suggesting is that we just don't provide a `matrix_exclude` at all for the unit test workflow and just use whatever that action defaults to. I don't think we can do the same yet for sanity because of ansible 2.9.

.github/workflows/all_green_check.yml
```diff
@@ -0,0 +1,42 @@
...
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - labeled
+      - unlabeled
```
abikouo, line 14: this triggers are impacting `changelog` workflow only, we should create a different file for that and all_green does not need to changelog being success as this will fail automatically
anyway we can keep it like that as it is already like this in the main branch

abikouo (approved):

PR #1748, opened 2023-09-07
Update EC2 Instance Tags Based on Conditions

theputta:
##### SUMMARY
PR includes condition to update the tags only if the existing Name tag is set to "LinuxOS."

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
amazon.aws.ec2_tag

### DIFF
--- a/plugins/modules/ec2_tag.py
+++ b/plugins/modules/ec2_tag.py
@@ -61,6 +61,22 @@ EXAMPLES = r"""
       Name: ubervol
       env: prod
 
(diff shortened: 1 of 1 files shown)


tremble (changes requested):
This example won't work at all.

a) In Ansible tags are (or should be) handled as a simple key/value pair, not the boto3 style list-of-dict
b) There is no condition parameter for ec2_tag

PR #1809, opened 2023-10-16
Add rds_global_cluster_info module

### LINKED ISSUE #1804: Add support for describe_global_clusters command to rds_cluster_info module
### Summary

The rds_cluster_info module currently provides support for retrieving information about RDS database clusters using the describe_db_clusters CLI command. To obtain information about global clusters, such as Aurora-MySQL, we will need to incorporate the describe_global_clusters CLI command into this module.

### Issue Type

Feature Idea

### Component Name

rds_cluster_info

### Additional Information

```yaml (paste below)

```

### Code of Conduct

GomathiselviS:
##### SUMMARY

Fixes #1804 
This PR adds rds_global_cluster_info module.

##### ISSUE TYPE

- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/rds_global_cluster_info.yml
+++ b/changelogs/fragments/rds_global_cluster_info.yml
@@ -1,3 +1,3 @@
 ---
-minor_changes::
-- Add rds_global_cluster_info module to get information Aurora gloabl db clusters.
+minor_changes:
+  - Add rds_global_cluster_info module to get information Aurora gloabl db clusters.
--- a/changelogs/fragments/rds_global_cluster_info.yml
+++ b/changelogs/fragments/rds_global_cluster_info.yml
@@ -1,3 +1,0 @@
----
-minor_changes:
-  - Add rds_global_cluster_info module to get information Aurora gloabl db clusters.
--- a/changelogs/fragments/rds_global_cluster_info.yml
+++ b/changelogs/fragments/rds_global_cluster_info.yml
@@ -1,0 +1,3 @@
+---
+minor_changes::
+- Add rds_global_cluster_info module to get information Aurora gloabl db clusters.
--- a/plugins/modules/rds_global_cluster_info.py
+++ b/plugins/modules/rds_global_cluster_info.py
@@ -22,6 +22,8 @@ options:
 
 author:
   - Gomathi Selvi Srinivasan (@GomathiselviS)
+notes:
+  - While developing this module, describe_global_cluster CLI did not yield any tag information. Consequently, the "tags" parameter is not included in this module.
 extends_documentation_fragment:
   - amazon.aws.common.modules
   - amazon.aws.region.modules
--- a/plugins/modules/rds_global_cluster_info.py
+++ b/plugins/modules/rds_global_cluster_info.py
@@ -146,7 +146,6 @@ from ansible.module_utils.common.dict_transformations import camel_dict_to_snake
 from ansible_collections.amazon.aws.plugins.module_utils.botocore import is_boto3_error_code
 from ansible_collections.amazon.aws.plugins.module_utils.modules import AnsibleAWSModule
 from ansible_collections.amazon.aws.plugins.module_utils.retries import AWSRetry
-from ansible_collections.amazon.aws.plugins.module_utils.transformation import ansible_dict_to_boto3_filter_list
 
 
 @AWSRetry.jittered_backoff(retries=10)
--- a/plugins/modules/rds_global_cluster_info.py
+++ b/plugins/modules/rds_global_cluster_info.py
@@ -1,0 +1,207 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2023 Ansible Project
+# Copyright (c) 2023 Gomathi Selvi Srinivasan (@GomathiselviS)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r"""
+module: rds_global_cluster_info
+version_added: 6.6.0
+short_description: Obtain information about Aurora global database clusters
+description:
+  - Obtain information about Aurora global database clusters.
+options:
+    global_cluster_identifier:
+        description:
+          - The user-supplied Global DB cluster identifier.
+          - If this parameter is specified, information from only the specific DB cluster is returned.
+          - This parameter isn’t case-sensitive.
+          - If supplied, must match an existing DBClusterIdentifier.
+        type: str
+    filters:
+        description:
+            - A filter that specifies one or more DB clusters to describe.
+              See U(https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_Filter.html).
+        type: dict
+author:
+  - Gomathi Selvi Srinivasan (@GomathiselviS)
+extends_documentation_fragment:
+  - amazon.aws.common.modules
+  - amazon.aws.region.modules
+  - amazon.aws.boto3
+"""
+
+EXAMPLES = r"""
+- name: Get info of all existing DB clusters
+  amazon.aws.rds_global_cluster_info:
+  register: _result_cluster_info
+
+- name: Get info on a specific DB cluster
+  amazon.aws.rds_global_cluster_info:
+    global_cluster_identifier: "{{ cluster_id }}"
+  register: _result_global_cluster_info
+
+- name: Get info all DB clusters with specific engine
+  amazon.aws.rds_cluster_info:
+    filters:
+      engine: "aurora-mysql"
+  register: _result_global_cluster_info
+"""
+
+RETURN = r"""
+global_clusters:
+  description: List of global clusters.
+  returned: always
+  type: list
+  contains:
+    global_cluster_identifier:
+        description: User-supplied global database cluster identifier..
+        type: str
+        sample: "ansible-test-global-cluster"
+    global_cluster_resource_id:
+        description:
+        - The Amazon Web Services Region-unique, immutable identifier for the global database cluster.
+        type: str
+        sample: cluster-xxx
+    global_cluster_arn:
+        description:
+        - The Amazon Resource Name (ARN) for the global database cluster.
+        type: str
+        sample: "arn:aws:rds::xxx:global-cluster:ansible-test-global-cluster"
+    status:
+        description: The status of the DB cluster.
+        type: str
+        sample: available
+    engine:
+        description: The database engine of the DB cluster.
+        type: str
+        sample: aurora-postgresql
+    engine_version:
+        description: The database engine version.
+        type: str
+        sample: 14.8
+    storage_encrypted:
+        description: Whether the DB cluster is storage encrypted.
+        type: bool
+        sample: false
+    deletion_protection:
+        description:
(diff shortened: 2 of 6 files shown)


changelogs/fragments/rds_global_cluster_info.yml
```diff
@@ -0,0 +1,3 @@
+---
```
alinabuzachis, line 1: No changelog fragment is needed for new modules. You just need to add skip-changelog label.

plugins/modules/rds_global_cluster_info.py
```diff
@@ -0,0 +1,210 @@
...
+"""
+
+RETURN = r"""
+global_clusters:
+  description: List of global clusters.
+  returned: always
+  type: list
+  contains:
+    global_cluster_identifier:
+        description: User-supplied global database cluster identifier..
```
alinabuzachis, line 59:
```suggestion
        description: User-supplied global database cluster identifier.
```

plugins/modules/rds_global_cluster_info.py
```diff
@@ -0,0 +1,210 @@
...
+    failover_state:
+        description:
+        - A data object containing all properties for the current state of an in-process or
+          pending switchover or failover process for this global cluster (Aurora global database).
+        - This object is empty unless the SwitchoverGlobalCluster or FailoverGlobalCluster operation was called on this global cluster.
+        type: dict
+        contains:
+            status:
+                description:
+                - The current status of the global cluster. Possible values are as follows
```
alinabuzachis, line 128: If this a sentence fragment or need to removed/completed?
```suggestion
                - The current status of the global cluster. 
```

plugins/modules/rds_global_cluster_info.py
```diff
@@ -0,0 +1,210 @@
...
+        module.fail_json_aws(e, "Couldn't get Global cluster information.")
+
+    return dict(
+        changed=False, global_clusters=[camel_dict_to_snake_dict(cluster, ignore_list=["Tags"]) for cluster in result]
+    )
+
+
+def main():
+    argument_spec = dict(
+        global_cluster_identifier=dict(),
```
alinabuzachis, line 192: I guess you can specify the type here.

tests/integration/targets/rds_global_cluster_create/tasks/main.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+    - name: Assert that the primary and secondary clusters are members of the global cluster
+      ansible.builtin.assert:
+        that:
+          - global_cluster_info.global_clusters[0].global_cluster_members[0].db_cluster_arn == primary_cluster_info_result.clusters[0].db_cluster_arn
+          - global_cluster_info.global_clusters[0].global_cluster_members[1].db_cluster_arn == secondary_cluster_info_result.clusters[0].db_cluster_arn
+          - global_cluster_info.global_clusters[0].engine == engine
+
+  always:
+    - name: Delete secondary cluster without creating a final snapshot
```
alinabuzachis, line 75: The rds_global_cluster must also be deleted.

tests/integration/targets/rds_global_cluster_create/tasks/main.yaml
```diff
@@ -0,0 +1,109 @@
...
+    - name: Create a read replica cluster for global database
+      amazon.aws.rds_cluster:
+        db_cluster_identifier: "{{ secondary_cluster_id }}"
+        region: "eu-north-1"
+        engine: "{{ engine }}"
+        engine_version: "{{ engine_version }}"
+        global_cluster_identifier: "{{ global_cluster_id }}"
+      register: replica_cluster
+
+    - name: Get Global DB information
```
alinabuzachis, line 49: Can you also please another task using filters rather than  global_cluster_identifier and some asserts?

plugins/modules/rds_global_cluster_info.py
GomathiselviS: Since the default value for "type" is "str," it can be left unpopulated.

tests/integration/targets/rds_global_cluster_create/tasks/main.yaml
GomathiselviS: It is deleted here - https://github.com/ansible-collections/amazon.aws/pull/1809/files/077fb7f396711db2c438f2e8d48971abf354418c#diff-cdb7a112dd4080b4c6c1ca6ee7c510392c5284863e6a47f2de1e1c48e8f8e2b7R103

tests/integration/targets/rds_global_cluster_create/tasks/main.yaml
```diff
@@ -0,0 +1,109 @@
...
+    - name: Delete primary cluster without creating a nal snapshot
+      amazon.aws.rds_cluster:
+        cluster_id: "{{ primary_cluster_id }}"
+        lobal_cluster_identifier: "{{  obal_cluster_id }}"
+        skip_final_snapshot: True
+        region: "{{ aws_region }}"
+        state: absent
+      ignore_errors: true
+
+    - name: Delete DB cluster without creating a final snapshot
```
alinabuzachis, line 103: Since it's a global rds cluster, shouldn't you use rds_global_cluster?

GomathiselviS: Fixed.

plugins/modules/rds_global_cluster_info.py
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2023 Ansible Project
+# Copyright (c) 2023 Gomathi Selvi Srinivasan (@GomathiselviS)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r"""
+module: rds_global_cluster_info
+version_added: 6.6.0
```
abikouo, line 10: Should it be `7.0.0`?

plugins/modules/rds_global_cluster_info.py
```diff
@@ -0,0 +1,210 @@
...
+  - Obtain information about Aurora global database clusters.
+options:
+    global_cluster_identifier:
+        description:
+          - The user-supplied Global DB cluster identifier.
+          - If this parameter is specified, information from only the specific DB cluster is returned.
+          - This parameter is not case-sensitive.
+          - If supplied, must match an existing DBClusterIdentifier.
+        type: str
+    filters:
```
abikouo, line 22: According to documentation here (https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeGlobalClusters.html) this parameter isn't currently supported, this can be removed and the modules restricted to a lister of cluster id

plugins/modules/rds_global_cluster_info.py
```diff
@@ -0,0 +1,210 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r"""
+module: rds_global_cluster_info
+version_added: 6.6.0
+short_description: Obtain information about Aurora global database clusters
+description:
+  - Obtain information about Aurora global database clusters.
+options:
+    global_cluster_identifier:
```
abikouo, line 15: This can be turned to a list

plugins/modules/rds_global_cluster_info.py
```diff
@@ -0,0 +1,210 @@
...
+
+- name: Get info all DB clusters with specific engine
+  amazon.aws.rds_cluster_info:
+    filters:
+      engine: "aurora-mysql"
+  register: _result_global_cluster_info
+"""
+
+RETURN = r"""
+global_clusters:
```
abikouo, line 53: if you keep the input identifier as a single string, you should set the output as a list of clusters info

abikouo (changes requested):

plugins/modules/rds_global_cluster_info.py
GomathiselviS: Yes, I think so. The next release should be 7.0.0 with all the migration PRs. Am I right?

plugins/modules/rds_global_cluster_info.py
GomathiselviS: I donot understand this comment. The argument `global_cluster_identifier` is defined as a string, while the return value is stored in a list named `global_clusters`. This aligns with the documentation provided by AWS Boto3 and is consistent with the existing RDS info module we have in place.

plugins/modules/rds_global_cluster_info.py
GomathiselviS: Why do you think so? The API accepts only a string

hakbailey (changes requested):
Just a few minor changes, otherwise this looks good.

hakbailey (approved):

plugins/modules/rds_global_cluster_info.py
```diff
@@ -0,0 +1,197 @@
...
+
+    params = dict()
+    if global_cluster_id:
+        params["GlobalClusterIdentifier"] = global_cluster_id
+
+    try:
+        result = _describe_global_clusters(client, **params)
+    except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
+        module.fail_json_aws(e, "Couldn't get Global cluster information.")
+
```
alinabuzachis, line 172: To be consistent with the other modules we should ensure tags parameter is always present in the returned data. However, describe_global_clusters does not return any Tags information, you should use https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds/client/list_tags_for_resource.html#list-tags-for-resource. You might be able to use this https://github.com/ansible-collections/amazon.aws/blob/main/plugins/module_utils/rds.py#L357

plugins/modules/rds_global_cluster_info.py
```diff
@@ -0,0 +1,197 @@
...
+            to_db_cluster_arn:
+                description: The Amazon Resource Name (ARN) of the Aurora DB cluster that is currently being promoted, and which is associated with this state.
+                type: str
+                sample: arn:aws:rds:us-east-2:123456789012:cluster:ansible-test-secondary
+            is_data_loss_allowed:
+                description:
+                - Indicates whether the operation is a global switchover or a global failover.
+                - If data loss is allowed, then the operation is a global failover. Otherwise, it is a switchover.
+                type: bool
+                sample: false
```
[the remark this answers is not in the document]
GomathiselviS: Aurora global clusters are not associated with any tags

plugins/modules/rds_global_cluster_info.py
GomathiselviS: Since no tags are linked to the global cluster, this module will not provide any tag information. I will make sure to include a note about this in the documentation for clarity.

alinabuzachis (approved):

plugins/modules/rds_global_cluster_info.py
abikouo: yes it accepts only a string, having a list is to avoid multiple calls on user side with different values of the identifier increasing the response time as for each one ansible need to build the env and the module needs to create a new session. But we can keep it like this for now

abikouo (approved):

PR #1815, opened 2023-10-19
Update README with Ansible version 2.13 as the minimum supported version

GomathiselviS:
##### SUMMARY

The documentation for amazon.aws is updated with the current state of supported Ansible versions. 

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,6 @@ AWS related modules and plugins supported by the Ansible community are in the [c
 
 Tested with the Ansible Core >= 2.13.0 versions, and the current development version of Ansible. Ansible Core versions prior to 2.13.0 are not supported.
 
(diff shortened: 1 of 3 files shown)


tremble (approved):
I'm good with this change.  I'd prefer that there's also a review from the cloud content team before merging

tremble (changes requested):
If we're dropping support for 2.12 We should also update runtime.yml:
https://github.com/ansible-collections/amazon.aws/blob/main/meta/runtime.yml#L2

alinabuzachis (commented):
https://github.com/ansible-collections/amazon.aws/blob/main/meta/runtime.yml#L2 Must also be updated.

tremble (approved):

alinabuzachis (approved):

PR #1919, opened 2023-12-11
s3_bucket_info - add support for bucket versioning in bucket_facts

abikouo:
##### SUMMARY
Module `s3_bucket_info` should be able to return the bucket versioning into bucket_facts

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`s3_bucket_info`

##### ADDITIONAL INFORMATION

```yaml
- amazon.aws.s3_bucket_info:
   name: mybucket
   bucket_facts:
    bucket_versioninig: true
```

expecting output

```json
{
    "bucket_name": "mybucket",
    "buckets": [
         {
            "bucket_versioning": {"Status": "Enabled"},    ## New information in output
           "creation_date": "2023-12-11T15:48:42+00:00",
           "name": "mybucket"
         }
    ],
   "changed": false,
   "msg": "Retrieved s3 info."
}
```

### DIFF
--- a/plugins/modules/s3_bucket_info.py
+++ b/plugins/modules/s3_bucket_info.py
@@ -36,70 +36,77 @@ options:
       - You can limit buckets by using the I(name) or I(name_filter) option.
     suboptions:
       bucket_accelerate_configuration:
-        description: Retrive S3 accelerate configuration.
+        description: Retrieve S3 accelerate configuration.
         type: bool
         default: False
       bucket_location:
-        description: Retrive S3 bucket location.
+        description: Retrieve S3 bucket location.
         type: bool
         default: False
       bucket_replication:
-        description: Retrive S3 bucket replication.
+        description: Retrieve S3 bucket replication.
         type: bool
         default: False
       bucket_acl:
-        description: Retrive S3 bucket ACLs.
+        description: Retrieve S3 bucket ACLs.
         type: bool
         default: False
       bucket_logging:
-        description: Retrive S3 bucket logging.
+        description: Retrieve S3 bucket logging.
         type: bool
         default: False
       bucket_request_payment:
-        description: Retrive S3 bucket request payment.
+        description: Retrieve S3 bucket request payment.
         type: bool
         default: False
       bucket_tagging:
-        description: Retrive S3 bucket tagging.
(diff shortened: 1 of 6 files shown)


plugins/modules/s3_bucket_info.py
```diff
@@ -396,6 +403,15 @@
               returned: always
               type: str
               sample: https
+    bucket_versioning:
+      description:
+        - The versioning state of the bucket.
+        - This will also specifies whether MFA delete is enabled in the bucket versioning configuration.
```
GomathiselviS, line 409:
```suggestion
        - This will also specify whether MFA delete is enabled in the bucket versioning configuration.
```

tests/integration/targets/s3_bucket_info/tasks/basic.yml
```diff
@@ -64,6 +66,8 @@
       - item.bucket_tagging["lowercase spaced"] == 'hello cruel world'
       - item.bucket_tagging["Title Case"] == 'Hello Cruel World'
       - item.bucket_location.LocationConstraint == aws_region
+      - item.bucket_versioning.Status is defined
```
GomathiselviS, line 69: Should we check for the value?

abikouo: @GomathiselviS for this specific test case I am just checking if the user request for the versioning information it is part of the ansible response, I don't want to check whether the versioning is enabled or not

GomathiselviS: Could you please clarify why not? The task's name indicates "Assert that retrieved bucket facts contain valid data," and it verifies the values of other parameters. Wouldn't it be appropriate to also verify the value of `bucket_versioning` parameter?

hakbailey (approved):

tests/integration/targets/s3_bucket_info/tasks/basic.yml
abikouo: @GomathiselviS I have added an extra validation to ensure that the versioning is enabled

GomathiselviS (approved):

PR #1979, opened 2024-02-20
Don't ignore IPv6 addresses when a single network interface is specified

maggu:
##### SUMMARY

When "ipv6_adresses" are used with "network" without specifying "interfaces", the parameter is ignored and no IPv6 adresses get set. I don't believe this is the intended or correct behaviour. 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ec2_instance

##### ADDITIONAL INFORMATION

```paste below

```


alinabuzachis:
@maggu Thanks for taking the time to submit this PR. Can you please add a changelog fragment https://docs.ansible.com/ansible/latest/community/development_process.html#changelogs-how-to?
To help avoid regressions, please cover this change with integration tests.

maggu:
@alinabuzachis Thanks for the reply and feedback. Regarding integration tests, there appear to currently be no tests at all for checking ec2_instance network settings.

What do you think would be the most appropriate approach for this? Adding it to ec2_instance_instance_*? Creating a new ec2_instance_network to test network settings? Something else, like ipv6_tests to test ivp6 settings across multiple modules?

maggu:
@alinabuzachis Changelog fragment added, and rebased to current main.

I believe that integration tests might need a wider scope than this PR, considering that currently similar functionality isn't tested and there's no obvious place to add it. Such tests would obviously be beneficial though.

hakbailey (commented):
I agree there is more work to do on this part of the module (including adding tests for the network interface options) and have added an [issue](https://github.com/ansible-collections/amazon.aws/issues/2054) for that, but for now this seems like a good bugfix.

hakbailey (approved):

abikouo (approved):

alinabuzachis (approved):

GomathiselviS:
@maggu If you can rebase this PR, we can merge it.

abikouo:
regate

PR #1992, opened 2024-02-23
Remove unneeded sanity ignore for ansible 2.17

hakbailey:
##### SUMMARY
Now that https://github.com/ansible/ansible/pull/82355 has been merged into devel we can remove the sanity ignore for 2.17 that has been causing non-blocking sanity tests to fail in CI.


hakbailey:
Alas, it hasn't been merged into milestone. Will re-submit when merged.

tremble (approved):
milestone's been updated!

PR #2083, opened 2024-05-10
backup_plan_info: bugfix to enable getting info of all backup plans

mandar242:
##### SUMMARY

Fix being unable to fetch info of all backup plans.
With `backup_plan_names` being a required parameter, the functionality to get all plans info was not working
```
# Gather information about all backup plans
- name: Get info of all backup plans 
  amazon.aws.backup_plan_info:
  register: plan_info_result

gave
**********
fatal: [localhost]: FAILED! => {"changed": false, "msg": "missing required arguments: backup_plan_names"}

```

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

backup_plan_info
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/2083-backup_plan_info-bugfix-get-info-for-all-plans.yml
+++ b/changelogs/fragments/2083-backup_plan_info-bugfix-get-info-for-all-plans.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - backup_plan_info - Bugfix to enable getting info of all backup plans (https://github.com/ansible-collections/amazon.aws/pull/2083).
--- a/plugins/modules/backup_plan_info.py
+++ b/plugins/modules/backup_plan_info.py
@@ -20,7 +20,6 @@ options:
   backup_plan_names:
     type: list
     elements: str
-    required: false
     description:
       - Specifies a list of plan names.
 extends_documentation_fragment:
@@ -122,8 +121,8 @@ def get_backup_plan_detail(client, module):
     if backup_plan_names is None:
         backup_plan_names = []
         backup_plan_list_info = get_all_backup_plans_info(client)["BackupPlansList"]
-        for mylist in backup_plan_list_info:
-            backup_plan_names.append(mylist["BackupPlanName"])
+        for backup_plan in backup_plan_list_info:
+            backup_plan_names.append(backup_plan["BackupPlanName"])
 
     for name in backup_plan_names:
         backup_plan_list.extend(get_plan_details(module, client, name))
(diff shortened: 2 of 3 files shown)


plugins/modules/backup_plan_info.py
```diff
@@ -20,7 +20,7 @@
   backup_plan_names:
     type: list
     elements: str
-    required: true
+    required: false
```
GomathiselviS, line 23: nitpick: `false` is the default value for `required`. So this line can be avoided.

plugins/modules/backup_plan_info.py
```diff
@@ -110,10 +110,21 @@
...
+
+
 def get_backup_plan_detail(client, module):
     backup_plan_list = []
     backup_plan_names = module.params.get("backup_plan_names")
 
+    if backup_plan_names is None:
+        backup_plan_names = []
+        backup_plan_list_info = get_all_backup_plans_info(client)["BackupPlansList"]
+        for mylist in backup_plan_list_info:
```
GomathiselviS, line 125: Could we use a variable name specific to the module rather than a generic or unrelated name like 'mylist'?

GomathiselviS (changes requested):
Can you add an example and a test to cover the changes made?

plugins/modules/backup_plan_info.py
```diff
@@ -122,7 +133,7 @@ def get_backup_plan_detail(client, module):
 
 def main():
     argument_spec = dict(
-        backup_plan_names=dict(type="list", elements="str", required=True),
+        backup_plan_names=dict(type="list", elements="str", required=False),
```
GomathiselviS, line 136:
```suggestion
        backup_plan_names=dict(type="list", elements="str"),
```

changelogs/fragments/2083-backup_plan_info-bugfix-get-info-for-all-plans.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
```
alinabuzachis, line 2: I don't know if this is a bugfix. This feature was not working because it was not implemented and the module does not say it supports listing all backup plans (that's why  `backup_plan_names` was a required parameter). To me it is probably minor_changes, a new functionality. You should also document that when O(backup_plan_names) is omitted the module gathers information about all backup plans.

alinabuzachis (approved):

tremble (approved):
I think it's fair to label this as a "bugfix" since the docs say it should have worked.

PR #2136, opened 2024-06-17
route53: Return diff only when module._diff is true

GomathiselviS:
##### SUMMARY

Refer https://issues.redhat.com/browse/ACA-1477
Route53 module, returns `after` and `before` diff as module output. This PR fixes this bug, by adding a check to return the diff only when module._diff is set to true.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/fix_module_diff_route53.yml
+++ b/changelogs/fragments/fix_module_diff_route53.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - modify the return value of the route53 module to return diff only when ``module._diff`` is set to true.
--- a/changelogs/fragments/fix_module_diff_route53.yml
+++ b/changelogs/fragments/fix_module_diff_route53.yml
@@ -1,2 +1,2 @@
 bugfixes:
-  - modify the return value of the route53 module to return diff only when ``module._diff`` is set to true.
+  - route53 - modify the return value to return diff only when ``module._diff`` is set to true (https://github.com/ansible-collections/amazon.aws/pull/2136).
--- a/plugins/modules/route53.py
+++ b/plugins/modules/route53.py
@@ -164,6 +164,15 @@ extends_documentation_fragment:
 """
 
 RETURN = r"""
+nameservers:
+  description: Nameservers associated with the zone.
+  returned: when state is 'get'
+  type: list
+  sample:
+  - ns-1036.awsdns-00.org.
+  - ns-516.awsdns-00.net.
+  - ns-1504.awsdns-00.co.uk.
+  - ns-1.awsdns-00.com.
 resource_record_sets:
   description: Info specific to the resource record.
   returned: when state is 'get'
--- a/plugins/modules/route53.py
+++ b/plugins/modules/route53.py
@@ -823,9 +823,9 @@ def main():
             diff=dict(
                 before=formatted_aws,
                 after=camel_dict_to_snake_dict(formatted_record) if command_in != "delete" else {},
-                resource_record_sets=rr_sets,
-            ),
-        )
+		),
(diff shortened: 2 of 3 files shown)


changelogs/fragments/fix_module_diff_route53.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - modify the return value of the route53 module to return diff only when ``module._diff`` is set to true.
```
abikouo, line 2:
```suggestion
  - route53 - modify the return value to return diff only when ``module._diff`` is set to true (https://github.com/ansible-collections/amazon.aws/pull/2136).
```

plugins/modules/route53.py
```diff
@@ -806,15 +816,20 @@ def main():
...
+    if module._diff:
+        module.exit_json(
+            changed=True,
+            wait_id=wait_id,
+            diff=dict(
+                before=formatted_aws,
+                after=camel_dict_to_snake_dict(formatted_record) if command_in != "delete" else {},
+                resource_record_sets=rr_sets,
+            ),
+        )
```
abikouo, line 828:
```suggestion
        module.exit_json(
            changed=True,
            wait_id=wait_id,
            diff=dict(
                before=formatted_aws,
                after=camel_dict_to_snake_dict(formatted_record) if command_in != "delete" else {},
            ),
            resource_record_sets=rr_sets,
        )
```

plugins/modules/route53.py
```diff
@@ -806,15 +816,20 @@ def main():
...
     module.exit_json(
         changed=True,
         wait_id=wait_id,
-        diff=dict(
-            before=formatted_aws,
-            after=formatted_record if command_in != "delete" else {},
-            resource_record_sets=rr_sets,
-        ),
+        resource_record_sets=[camel_dict_to_snake_dict(formatted_record)] if command_in != "delete" else {},
```
abikouo, line 832: the returned value for `resource_record_sets` should be the same whether we are running in diff mode or not

abikouo (changes requested):

plugins/modules/route53.py
```diff
@@ -806,15 +816,20 @@ def main():
...
-
+    if module._diff:
+        module.exit_json(
+            changed=True,
+            wait_id=wait_id,
+            diff=dict(
+                before=formatted_aws,
+                after=camel_dict_to_snake_dict(formatted_record) if command_in != "delete" else {},
+		),
+        	resource_record_sets=[camel_dict_to_snake_dict(formatted_record)] if command_in != "delete" else {},
```
abikouo, line 827: this is should not be part of the diff structure

abikouo: to avoid duplicate code, you can do something like

```
result = dict(changed=True, wait_id=wait_id, resource_record_sets=[...])
if module._diff:
    result.update(dict(before=formatted_aws, after=....))
module.exit_json(**result)
```

abikouo (changes requested):

abikouo (approved):

hakbailey (approved):

GomathiselviS:
regate

GomathiselviS:
regate

plugins/modules/route53.py
```diff
@@ -203,6 +203,11 @@
       description: An identifier that differentiates among multiple resource record sets that have the same combination of DNS name and type.
       returned: always
       type: str
+    name:
+      description: Domain name for the record set.
+      returned: always
+      type: str
+      sample: new.foo.com.
```
alinabuzachis, line 210:
```suggestion
      sample: "new.foo.com"
```

PR #2153, opened 2024-06-27
A new workflow to ease pull request review by labelling them

abikouo:
This new workflow calculates the size of a pull request based on the additions and deletions and add a label corresponding to the size of the pull (XS, S, M, L, XL).
This aims to ease pull request review
- Prior reviewing the PR, reviewers can estimate the time to spend on it according to the size
- PR authors should consider this information to split the PR to ease the review

### DIFF
--- a/.github/actions/pull_request_size/check_pr.py
+++ b/.github/actions/pull_request_size/check_pr.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 # (c) 2024 Aubin Bikouo <@abikouo>
 # GNU General Public License v3.0+
@@ -34,9 +33,7 @@ def WriteComment(repository: str, pr_number: int, comment: str) -> None:
     )
     # Successful call to the API will return '201' (created)
     if result.status_code != 201:
-        raise RuntimeError(
-            f"Post to URL {url} returned status code = {result.status_code}"
-        )
+        raise RuntimeError(f"Post to URL {url} returned status code = {result.status_code}")
 
 
 def AddLabelToPR(repository: str, pr_number: int, type: str) -> None:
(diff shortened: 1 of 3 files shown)


GomathiselviS:
Hi @abikouo 

Thank you for the effort you put into creating a GH workflow to ease PR review by labeling the PR based on its size. This is a fantastic step towards improving our review process.

However, I would like to note that while this workflow can inform the PR submitter about the need to break down large PRs, we should be cautious about relying solely on the formula used. The PR size should ideally reflect the complexity of the inserted and deleted lines, not just their count. The current formula may not fully capture this complexity.

Thank you again for your contribution. Your work is much appreciated, and this workflow will certainly help streamline our PR submission and review process. Can you add a proposal in the cloud-content-handbook and move the GH action to [cloud-content-ci-automation repository](https://github.com/ansible/cloud-content-ci-automation)?

.github/actions/pull_request_size/check_pr.py
```diff
@@ -0,0 +1,144 @@
...
+    response = requests.put(url_base + "labels", headers=headers, json={"labels": [f"size/{type}"]})
+    if response.status_code != 200:
+        raise RuntimeError(
+            f"Unable to add label '{label}' to issue {repository}/{pr_number} - status_code = {response.status_code}"
+        )
+
+
+def LabelCommentPR(repository: str, pr_number: int, insertions: int, deletions: int) -> None:
+    # Calculating PR Size:
+    # The PR size is calculated using the formula: insertions + deletions * 0.5.
```
alinabuzachis, line 75: Where this formula is coming from?

alinabuzachis (commented):
As @GomathiselviS suggested, I think this approach should first be proposed in the content handbook.

PR #2165, opened 2024-07-01
Refactor ec2_eip* modules

abikouo:
##### SUMMARY
Refactor ec2_eip,ec2_eip_info modules

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ec2_eip ec2_eip_info


plugins/modules/ec2_eip_info.py
```diff
@@ -91,7 +91,7 @@
       type: str
       sample: "eipassoc-0fe9ce90d6e983e97"
     "domain":
-      description: The network (vpc).
+      description: The network ( vpc).
```
hakbailey, line 94:
```suggestion
      description: The network (vpc).
```

plugins/modules/ec2_eip_info.py
```diff
@@ -18,7 +18,7 @@
   filters:
     description:
       - A dict of filters to apply. Each dict item consists of a filter key and filter
-        value. See U(https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-addresses.html#options)
+        value.  See U(https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-addresses.html#options)
```
hakbailey, line 21:
```suggestion
        value. See U(https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-addresses.html#options)
```

plugins/modules/ec2_eip.py
```diff
@@ -18,28 +18,28 @@
     description:
       - The id of the device for the EIP.
       - Can be an EC2 Instance id or Elastic Network Interface (ENI) id.
-      - When specifying an ENI id, O(in_vpc) must be V(true).
```
hakbailey, line 21: These changes to the documentation block should be reverted, the semantic markup was correct.

plugins/modules/ec2_eip.py
```diff
@@ -234,13 +241,19 @@ class EipError(Exception):
...
+    allow_reassociation: bool,
+    is_instance: bool = True,
+) -> bool:
     if address_is_associated_with_device(ec2, module, address, device_id, is_instance):
-        return {"changed": False}
+        return False
 
     # If we're in check mode, nothing else to do
-    if not check_mode:
+    if not module.check_mode:
```
hakbailey, line 256: For readability, I would suggest maybe setting `address` to `{}` here if it's supplied as `None`, rather than having several `address or {}` checks below. Or better yet, don't make it optional and instead require the check outside this function. Same for a few other functions below.

plugins/modules/ec2_eip.py
```diff
@@ -525,7 +540,7 @@ def allocate_address_from_pool(
     public_ipv4_pool,
     tags,
 ):
-    # type: (EC2Connection, AnsibleAWSModule, str, bool, str) -> Address
+    # type: (Any, AnsibleAWSModule, str, bool, str, Dict) -> Optional[Dict]
```
hakbailey, line 543: Can this be a type hint instead of a comment?

plugins/modules/ec2_eip.py
abikouo: I forgot to rebase my PR

plugins/modules/ec2_eip.py
abikouo: updated.

plugins/modules/ec2_eip.py
```diff
@@ -249,110 +264,127 @@ def associate_ip_and_device(
...
+            address = {}
         try:
-            if address["Domain"] == "vpc":
-                ec2.disassociate_address(AssociationId=address["AssociationId"], aws_retry=True)
+            if address.get("Domain") == "vpc":
+                disassociate_address(ec2, association_id=address.get("AssociationId"))
             else:
-                ec2.disassociate_address(PublicIp=address["PublicIp"], aws_retry=True)
-        except (botocore.exceptions.BotoCoreError, botocore.exceptions.ClientError) as e:
+                # TODO: fix this, 'PublicIP' argument has been deprecated for function 'disassociate_address'
```
alinabuzachis, line 314: Can you please open an issue to keep track of this?

plugins/modules/ec2_eip.py
```diff
@@ -249,110 +264,127 @@ def associate_ip_and_device(
...
+    is_instance: bool = True,
+) -> bool:
     if not address_is_associated_with_device(ec2, module, address, device_id, is_instance):
-        return {"changed": False}
+        return False
 
     # If we're in check mode, nothing else to do
-    if not check_mode:
+    if not module.check_mode:
+        if not address:
```
alinabuzachis, line 308: What do you think about modifying line 632 to something like?

       if device_id:
            address = find_address(ec2, module, public_ip, device_id, is_instance=is_instance) or {}
        else:
            address = find_address(ec2, module, public_ip, None) or {}

and limit these checks?
```
        if not address:
            address = {}
```

alinabuzachis (approved):

plugins/modules/ec2_eip.py
abikouo: I have done some more code refactoring, could you please review once again? Thanks

alinabuzachis (approved):

hakbailey (approved):

PR #2286, opened 2024-09-06
cloudwatchlogs_log_group_metric_filter: add support for unit and dimensions

jmisset-cb:
##### SUMMARY

This PR adds support for the options `unit` and `dimensions` in the `cloudwatchlogs_log_group_metric_filter` module.
This enables configuring `unit` and `dimensions` in Cloudwatch Logs Metricfilters using ansible, which was previously not possible.

The addition is pretty straigthforward since both `unit` and `dimensions` are part of the `metric_transformation` parameter.

`dimensions` and `default_value` are mutually exclusive, however:

1. The AWS API does not fail when both are present, instead the dimensions are simply ignored.
2. Since they are not top-level parameters, the ansible module option `mutually_exclusive` was not possible. Instead a custom check was added to the module that throws an error when both parameters are present.

An integration test has been added for this case, as well as for configuring metric_filters with units and/or dimensions.

The function metricTransformationHandler has been rewritten slightly due to the addition of the two extra optional parameters, to make it a bit more readable.

Happy to get your feedback!

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

cloudwatchlogs_log_group_metric_filter

##### ADDITIONAL INFORMATION

```paste below

```


plugins/modules/cloudwatchlogs_log_group_metric_filter.py
```diff
@@ -184,6 +215,12 @@ def main():
         metricTransformation = [camel_dict_to_snake_dict(item) for item in [originMetricTransformations]]
 
     elif state == "present":
+        if (
+            metric_transformation.get("default_value") is not None
+            and metric_transformation.get("dimensions") is not None
+        ):
+            module.fail_json(msg="default_value and dimensions are mutually exclusive.")
```
GomathiselviS, line 222: Since the AWS Boto3 call does not fail or throw an error when both dimensions and default values are provided, can we issue a warning and proceed with the configuration, rather than causing the module to fail?

GomathiselviS (changes requested):
Thank you for working on this PR. Please update the `Return` block of the module documentation with the new parameters.

plugins/modules/cloudwatchlogs_log_group_metric_filter.py
jmisset-cb: That is possible, however personally I prefer it to fail, because if this is only a warning, then:

1. I feel that does increase the risk that this goes unnoticed for the user, and can lead to confusion on why a metric_filter would not work as expected.
2. You will run into idempotency issues, because the dimensions will never actually be configured, so on subsequent runs tasks with both dimensions and default_value will always be 'changed'.

If you still feel this should be a warning instead of a failure, let me know then I will update the PR.

jmisset-cb: Thank you for the quick response by the way. I have updated the `Return` block

GomathiselviS: I would like to know  what @tremble , @alinabuzachis think about throwing a warning instead of an error for the above condition.

plugins/modules/cloudwatchlogs_log_group_metric_filter.py
```diff
@@ -56,7 +56,20 @@
         default_value:
           description:
             - The value to emit when a filter pattern does not match a log event.
+            - The I(default_value) and I(dimensions) options are mutually exclusive.
           type: float
+        unit:
+          description:
```
alinabuzachis, line 62: Can you please add `version_added: 8.3.0` for each of these two new options?

jmisset-cb: Done

alinabuzachis (approved):

GomathiselviS (approved):

alinabuzachis:
@jmisset-cb Can you please rebase this branch?

jmisset-cb:
> @jmisset-cb Can you please rebase this branch?

@alinabuzachis Done!

PR #2297, opened 2024-09-18
Fix sanity errors occurring with the devel branch of ansible

alinabuzachis:
##### SUMMARY

Fix sanity errors occurring with the devel branch of ansible 

```
ERROR: Found 2 pylint issue(s) which need to be resolved:
ERROR: plugins/modules/ec2_vpc_endpoint.py:441:24: used-before-assignment: Using variable 'result' before assignment
ERROR: plugins/modules/ec2_vpc_route_table.py:792:7: used-before-assignment: Using variable 'route_table' before assignment
```
See https://github.com/ansible-collections/amazon.aws/pull/2295

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/20240919-sanity_fixes_devel.yml
+++ b/changelogs/fragments/20240919-sanity_fixes_devel.yml
@@ -1,0 +1,2 @@
+trivial:
+  - "Fix sanity errors occurring with the devel branch of ansible."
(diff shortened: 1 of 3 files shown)


tremble (approved):
Just a side note: these are actually cropping up because the various static analysis tools don't know that fail_json_aws ultimately calls `exit()`.

As we move away from passing module around and using exceptions which we handle in the main() loop, these will start to go away...

Issue #1976, opened 2024-02-14
Allow Route53 DNSSEC Signing


rvgate:
### Summary

Route53 allows enabling DNSSEC keys on registered domain, and also enable DNSSEC signing per hosted zone. It would be amazing if this was configurable through Ansible as well.

It would probably be a combination of using kms module to create the key and then applying it to the zone.

kms_key: seems to be missing the action GetPublicKey to configure this correctly
route53_zone: allow DNSSEC to be enabled, referring to a key

the final step would be to enable the trust chain somehow, which is configured at the "Registered domains" section within AWS. I dont think there is currently an Ansible module for that.

### Issue Type

Feature Idea

### Component Name

route53_zone, route53, kms_key

### Additional Information

The feature would combine multiple modules to fully enable DNSSEC signing for a domain.

Step 1)
Configure an KMS key using the kms_key module, ending up with the required policy (principal/serviceaccount masked):

```json
{
    "Version": "2012-10-17",
    "Id": "dnssec-policy",
    "Statement": [
        {
            "Sid": "Enable IAM User Permissions",
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::XXX:YYY"
            },
            "Action": "kms:*",
            "Resource": "*"
        },
        {
            "Sid": "Allow Route 53 DNSSEC Service",
            "Effect": "Allow",
            "Principal": {
                "Service": "dnssec-route53.amazonaws.com"
            },
            "Action": [
                "kms:DescribeKey",
                "kms:GetPublicKey",
                "kms:Sign"
            ],
            "Resource": "*",
            "Condition": {
                "StringEquals": {
                    "aws:SourceAccount": "XXX"
                },
                "ArnLike": {
                    "aws:SourceArn": "arn:aws:route53:::hostedzone/*"
                }
            }
        },
        {
            "Sid": "Allow Route 53 DNSSEC to CreateGrant",
            "Effect": "Allow",
            "Principal": {
                "Service": "dnssec-route53.amazonaws.com"
            },
            "Action": "kms:CreateGrant",
            "Resource": "*",
            "Condition": {
                "Bool": {
                    "kms:GrantIsForAWSResource": "true"
                }
            }
        }
    ]
}
```

Step 2)
Configure the hosted zone, enabling DNSSEC signing:

```yaml
- name: create hosted zone
  amazon.aws.route53_zone:
    access_key: "{{ aws.access_key_id }}"
[...]

PR #2412, opened 2024-12-04
route53_key_signing_key -  new module

alinabuzachis:
##### SUMMARY

Terminator PR https://github.com/mattclay/aws-terminator/pull/312

Closes https://github.com/ansible-collections/amazon.aws/issues/1976

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

route53_key_signing_key

##### ADDITIONAL INFORMATION

```paste below

```


alinabuzachis:
recheck

GomathiselviS:
I would recommend creating a separate PR to add the new module for better clarity and easier review, rather than combining it with the addition of DNSSEC signing functionality.

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,360 @@
...
+options:
+    state:
+        description:
+            - Whether or not the zone should exist or not.
+        default: present
+        choices: [ "present", "absent" ]
+        type: str
+    caller_reference:
+        description:
+            - A unique string that identifies the request.
```
GomathiselviS, line 28: Required when state is present

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,360 @@
...
+        type: str
+    hosted_zone_id:
+        description:
+            - The unique string (ID) used to identify a hosted zone.
+        type: str
+        required: true
+        aliases: ["zone_id"]
+    key_management_service_arn:
+        description:
+            - The Amazon resource name (ARN) for a customer managed key in Key Management Service (KMS).
```
GomathiselviS, line 39: Add "Required when state is present"

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,360 @@
...
+    key_management_service_arn: "{{ kms_key.key_arn }}"
+    caller_reference: "{{ aws_caller_info.arn }}"
+    status: "ACTIVE"
+    state: present
+
+- name: Delete a Key Signing Key Request and deactivate it
+  amazon.aws.route53_ksk:
+    name: "{{ resource_prefix }}-ksk"
+    hosted_zone_id: "{{ _hosted_zone.zone_id }}"
+    status: "INACTIVE"
```
GomathiselviS, line 95: Is status needed when we delete?

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,361 @@
...
+
+    if ksk is not None:
+        response = {"KeySigningKey": ksk}
+        if ksk["Status"] != status:
+            changed = True
+
+            if module.check_mode:
+                module.exit_json(
+                    changed=changed,
+                    msg=f"Would have updated the Key Signing Key status to {status} if not in check_mode.",
```
GomathiselviS, line 269: Can we return the changed dict (would have changed if not check_mode = true) as well?

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,361 @@
...
+                )
+
+            if status == "ACTIVE":
+                response.update(activate(client, zone_id, name))
+            elif status == "INACTIVE":
+                response.update(deactivate(client, zone_id, name))
+    else:
+        changed = True
+        if module.check_mode:
+            module.exit_json(changed=changed, msg="Would have created the Key Signing Key if not in check_mode.")
```
GomathiselviS, line 279: Can we return the changed dict (would have changed if not check_mode = true) as well?

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,361 @@
...
+    name = module.params.get("name")
+    response = {"KeySigningRequest": {}}
+
+    if ksk is not None:
+        changed = True
+        response["KeySigningRequest"] = ksk
+        if module.check_mode:
+            module.exit_json(changed=changed, msg="Would have deleted the Key Signing Key if not in check_mode.")
+
+        if module.params.get("status") == "INACTIVE":
```
GomathiselviS, line 306: Is it possible for a condition to have state = "delete" and status = "ACTIVE"? I'm not sure if such a scenario can exist. Do we need this condition? We can ignore status for deletion. I think we should deactivate the key signing key before deleting in any case.

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,361 @@
...
+
+def delete(client, module: AnsibleAWSModule, ksk):
+    changed: bool = False
+    zone_id = module.params.get("hosted_zone_id")
+    name = module.params.get("name")
+    response = {"KeySigningRequest": {}}
+
+    if ksk is not None:
+        changed = True
+        response["KeySigningRequest"] = ksk
```
GomathiselviS, line 302: Is this needed? response is getting overwritten [here](https://github.com/ansible-collections/amazon.aws/pull/2412/files#diff-0de787964421af41ce7865d4ab961a54e9dfee7a5ea7a10cff5bcc01824aeb6eR311)

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+def create_or_update(client, module: AnsibleAWSModule, ksk):
+    changed: bool = False
+    zone_id = module.params.get("hosted_zone_id")
+    name = module.params.get("name")
+    status = module.params.get("status")
+
+    if ksk is not None:
+        response = {"KeySigningKey": ksk}
```
GomathiselviS, line 262: Is this part of the return dict?

tests/integration/targets/route53_ksk/tasks/main.yml
```diff
@@ -0,0 +1,258 @@
...
+        status: "ACTIVE"
+        state: present
+        wait: true
+      register: _ksk_request
+
+    - name: Assert success
+      ansible.builtin.assert:
+        that:
+          - _ksk_request is successful
+          - '"change_info" in _ksk_request'
```
GomathiselviS, line 119: can we check if the status is ACTIVE?

tests/integration/targets/route53_ksk/tasks/main.yml
```diff
@@ -0,0 +1,258 @@
...
+        key_management_service_arn: "{{ kms_key.key_arn }}"
+        caller_reference: "{{ aws_caller_info.arn }}"
+        status: "INACTIVE"
+        state: present
+      register: _ksk_request
+
+    - name: Assert success
+      ansible.builtin.assert:
+        that:
+          - _ksk_request is successful
```
GomathiselviS, line 151: Can we check the status here?

tests/integration/targets/route53_ksk/tasks/main.yml
```diff
@@ -0,0 +1,258 @@
...
+          - _ksk_request.changed
+          - '"msg" in _ksk_request'
+          - '"Would have deleted the Key Signing Key if not in check_mode." in _ksk_request.msg'
+
+    - name: Delete the Key Signing Key request
+      amazon.aws.route53_ksk:
+        name: "{{ resource_prefix }}-ksk"
+        hosted_zone_id: "{{ _hosted_zone.zone_id }}"
+        status: "INACTIVE"
```
GomathiselviS, line 192: I am not sure about "status: INACTIVE" for deletion.

alinabuzachis: status: "INACTIVE" is not for deletion. `state: absent` is for deletion. status: "INACTIVE" deactivate the ksk.

plugins/modules/route53_ksk.py
alinabuzachis: This situation was already handled because there was a note in the description asking the user to deactivate the key before deleting it. However, since it is not possible to delete a key without first deactivating it, eliminating an additional step or the need to specify the status: INACTIVE, I handled deactivation by default when the user wants to delete it.

tests/integration/targets/route53_ksk/tasks/main.yml
alinabuzachis: This is not possible because `get_change` returns the status of the request. There is no get_* functionality in the API that directly provides information about the KSK. I added a check for the status of the request instead.

tests/integration/targets/route53_ksk/tasks/main.yml
alinabuzachis: Same as above.

tests/integration/targets/route53_ksk/tasks/main.yml
alinabuzachis: No, we do not need it to delete a KSK. I updated the examples.

plugins/modules/route53_ksk.py
alinabuzachis: Removed this part since not needed.

plugins/modules/route53_ksk.py
GomathiselviS: Apologies for the confusion in my previous statement. What I meant was: in addition to the 'changed' status, do you think it's possible to also return the "response/return dictionary," even if no changes are made to the resource? By definition, any module designed to support 'check mode' reports what changes would have been made without actually applying them. However, this approach is not followed by our cloud modules. I was wondering if we could start implementing this behavior in new modules and then work on modifying the existing ones later.

alinabuzachis: No worries, understood, done!

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,366 @@
...
+    zone_id = module.params.get("hosted_zone_id")
+    name = module.params.get("name")
+    status = module.params.get("status")
+    response = {}
+
+    if ksk:
+        if ksk["Status"] != status:
+            changed = True
+
+            if module.check_mode:
```
GomathiselviS, line 264: ksk["Status"] = status

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,366 @@
...
+    zone_id = module.params.get("hosted_zone_id")
+    name = module.params.get("name")
+    response = {}
+
+    if ksk:
+        changed = True
+        if module.check_mode:
+            module.exit_json(
+                changed=changed,
+                key_signing_key=camel_dict_to_snake_dict(ksk),
```
GomathiselviS, line 302:
```suggestion
                key_signing_key={},
```

GomathiselviS (changes requested):

plugins/modules/route53_ksk.py
alinabuzachis: Why is this needed?

plugins/modules/route53_ksk.py
alinabuzachis: I thought we could return the info about the resource that will be deleted.

GomathiselviS: From what I understand, check mode allows you to simulate the changes that would be made by the playbook or task. It shows you what would happen if the playbook were run in a live environment, but without making any actual changes to the actual resources. This is useful when you want to see the impact of the changes beforehand. Therefore, when a delete operation is performed, the return dictionary will be empty, which is the expected behavior here.

plugins/modules/route53_ksk.py
GomathiselviS: For the same reason as the delete operation, when the resource is actually updated, the value of `ksk["Status"]` will be modified or set.

alinabuzachis: Okay, done although as mentioned below there is the informational message saying that the status would have been updated to ACTIVE/INACTIVE if it was not in control mode.
In general, I don't really know if  we tweak the retuned result of the resource in check mode to reflect the desired updates though. If we do so, the msg probably becomes useless... but I don't know.

plugins/modules/route53_ksk.py
GomathiselviS: Thanks for doing this! I wanted the cloud modules to adhere to the [definition](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_checkmode.html#using-check-mode) of check mode that is used consistently across Ansible.

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,367 @@
...
+description:
+    - Creates a new key-signing key (KSK) associated with a hosted zone.
+      You can only have two KSKs per hosted zone.
+    - When O(state=absent), it deactivates and deletes a key-signing key (KSK).
+    - Activates a key-signing key (KSK) so that it can be used for signing by DNSSEC.
+    - Deactivates a key-signing key (KSK) so that it will not be used for signing by DNSSEC.
+options:
+    state:
+        description:
+            - Whether or not the zone should exist or not.
```
abikouo, line 20:
```suggestion
            - Whether or not the zone should exist.
```

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,367 @@
...
+author:
+    - Alina Buzachis (@alinabuzachis)
+"""
+
+EXAMPLES = r"""
+- name: Create a Key Signing Key Request
+  amazon.aws.route53_ksk:
+    name: "{{ resource_prefix }}-ksk"
+    hosted_zone_id: "{{ _hosted_zone.zone_id }}"
+    key_management_service_arn: "{{ kms_key.key_arn }}"
```
abikouo, line 76:
```suggestion
    key_management_service_arn: "arn:aws:kms:us-west-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab"
```

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,367 @@
...
+    caller_reference: "{{ aws_caller_info.arn }}"
+    status: "INACTIVE"
+    state: present
+
+- name: Activate a Key Signing Key Request
+  amazon.aws.route53_ksk:
+    name: "{{ resource_prefix }}-ksk"
+    hosted_zone_id: "{{ _hosted_zone.zone_id }}"
+    key_management_service_arn: "{{ kms_key.key_arn }}"
```
abikouo, line 85:
```suggestion
    key_management_service_arn: "arn:aws:kms:us-west-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab"
```

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,367 @@
...
+    caller_reference: "{{ aws_caller_info.arn }}"
```
abikouo, line 86:
```suggestion
    caller_reference: "arn:aws:iam::123456789012:user/SomeUser"
```

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,367 @@
...
+    - Alina Buzachis (@alinabuzachis)
+"""
+
+EXAMPLES = r"""
+- name: Create a Key Signing Key Request
+  amazon.aws.route53_ksk:
+    name: "{{ resource_prefix }}-ksk"
+    hosted_zone_id: "{{ _hosted_zone.zone_id }}"
+    key_management_service_arn: "{{ kms_key.key_arn }}"
+    caller_reference: "{{ aws_caller_info.arn }}"
```
abikouo, line 77:
```suggestion
    caller_reference: "arn:aws:iam::123456789012:user/SomeUser"
```

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,367 @@
...
+    status: "INACTIVE"
+    state: present
+
+- name: Activate a Key Signing Key Request
+  amazon.aws.route53_ksk:
+    name: "{{ resource_prefix }}-ksk"
+    hosted_zone_id: "{{ _hosted_zone.zone_id }}"
```
abikouo, line 84:
```suggestion
    hosted_zone_id: "ZZZ1111112222"
```

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,367 @@
...
+    - amazon.aws.boto3
+author:
+    - Alina Buzachis (@alinabuzachis)
+"""
+
+EXAMPLES = r"""
+- name: Create a Key Signing Key Request
+  amazon.aws.route53_ksk:
+    name: "{{ resource_prefix }}-ksk"
+    hosted_zone_id: "{{ _hosted_zone.zone_id }}"
```
abikouo, line 75:
```suggestion
    hosted_zone_id: "ZZZ1111112222"
```

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,367 @@
...
+    key_management_service_arn: "{{ kms_key.key_arn }}"
+    caller_reference: "{{ aws_caller_info.arn }}"
+    status: "ACTIVE"
+    state: present
+
+- name: Delete a Key Signing Key Request and deactivate it
+  amazon.aws.route53_ksk:
+    name: "{{ resource_prefix }}-ksk"
+    hosted_zone_id: "{{ _hosted_zone.zone_id }}"
```
abikouo, line 93:
```suggestion
    hosted_zone_id: "ZZZ1111112222"
```

plugins/modules/route53_ksk.py
```diff
@@ -0,0 +1,367 @@
...
+        if module.check_mode:
+            module.exit_json(
+                changed=changed,
+                key_signing_key={},
+                msg="Would have deleted the Key Signing Key if not in check_mode.",
+            )
+
+        result = deactivate(client, zone_id, name)
+        change_id = result["ChangeInfo"]["Id"]
+        wait(client, module, change_id)
```
abikouo, line 309:
```suggestion
```
The ``wait`` operation is performed on line ``351``, this one is not conditional

abikouo (changes requested):
I prefer to rename the module to ``route53_key_signing_key``, having ``route53_ksk`` is not meaningful.
Can you also please add type hinting to all functions?

plugins/modules/route53_ksk.py
alinabuzachis: The wait is necessary to wait for the deactivation of the key signing key, which must be done before deleting it. The wait operation on line 351 waits for deletion when wait=true.

abikouo (approved):

PR #2462, opened 2025-01-14
Revert stable 8 backports

hakbailey:
##### SUMMARY
Reverts two backports for minor features that were intended to be released in 8.3.0, but were instead released in 9.0.0. As such, they are being removed from this branch to reflect what we actually did.


alinabuzachis (approved):
probably would have preferred two different PRs for each revert, but LGTM!

gravesm (approved):

tremble (approved):

PR #2472, opened 2025-01-16
Add Github worflow that adds backport_failed label when automatic backports fail

alinabuzachis:
##### SUMMARY

Add Github worflow that adds backport_failed label when automatic backports fail.
This will help keep track of any failed back doors that require manual actions to be taken.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/check_backports.yml
+++ b/.github/workflows/check_backports.yml
@@ -1,0 +1,31 @@
+name: Detect Patchback Bot Backport Failures
+
+on:
+  pull_request:
+    branches:
+      - main
+    types:
+      - closed
+
+jobs:
+  detect_backport_failure:
+    if: github.event.pull_request.merged == true  # Only run for merged PRs
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check for Patchback Bot Comments
+        id: check_comments
+        shell: bash
+        run: |
+          # Check if the comment exists from patchback with the specific failure text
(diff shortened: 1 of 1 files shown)


tremble (changes requested):
Because the PR will be closed before the comment's added, I *think* this isn't going to fire.

As an alternative, we could fire the workflow using
```
on:
  issue_comment:
    - edited
```

We should then be able to directly evaluate `${{ issue_comment.comment }}` and only look at the comment that patchback added.  (It's done as a three-step process, add comment, create PR, edit comment)

This would then mean that once manually resolved we can remove the backport_failed label and not have the action re-fire to relabel.

alinabuzachis:
> Because the PR will be closed before the comment's added, I _think_ this isn't going to fire.
> 
> As an alternative, we could fire the workflow using
> 
> ```
> on:
>   issue_comment:
>     - edited
> ```
> 
> We should then be able to directly evaluate `${{ issue_comment.comment }}` and only look at the comment that patchback added. (It's done as a three-step process, add comment, create PR, edit comment)
> 
> This would then mean that once manually resolved we can remove the backport_failed label and not have the action re-fire to relabel.

It is trigged, but the only thing is that it maybe won't wait enough for the patchback to add comment, create PR and edit comment. I'm updating it using your suggestion.

tremble (approved):

PR #2807, opened 2026-01-16
EC2 Indirect Node Count Query

chynasan:
##### SUMMARY

This work is relevant to https://issues.redhat.com/browse/ACA-4693.

Tested with https://play.jqlang.org.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
Indirect Node Count Query

### DIFF
--- a/extensions/audit/event_query.yml
+++ b/extensions/audit/event_query.yml
@@ -1,4 +1,20 @@
 ---
+amazon.aws.ec2_instance_info:
+  query: >-
+    .instances[] | select(. != null) | {
+      name: (.name // .tags.Name // "UNKNOWN"),  # unique identifier for the resource
+      canonical_facts: {
+      id: .instance_id,
+      status: .state.name,
+      tags: (.tags // {}),
+      },
+      facts: {
+        infra_type: "PublicCloud",
+        infra_bucket: "Compute",
+        device_type: "EC2"
+        },
+      }
+
 amazon.aws.s3_bucket_info:
   query: >-
     .buckets[] | select(. != null) |
--- a/tests/integration/targets/node_query_instance_ec2/play.yaml
+++ b/tests/integration/targets/node_query_instance_ec2/play.yaml
@@ -1,0 +1,66 @@
+- hosts: localhost
+  gather_facts: true
+
+  roles:
+    - role: setup_node_query
+
+  module_defaults:
+    group/aws:
+      access_key: "{{ aws_access_key }}"
+      secret_key: "{{ aws_secret_key }}"
+      session_token: "{{ security_token | default(omit) }}"
+      region: "{{ aws_region }}"
+
+  vars:
+    ec2_instance_buckets:
+      - name: "instance-{{ resource_prefix }}-1"
+        tags:
+          resource_prefix: "{{ resource_prefix }}"
+    ec2_instance_facts:
+      device_type: "EC2"
+      infra_bucket: "Compute"
+      infra_type: "PublicCloud"
+
+  tasks:
+    - block:
+        # EC2 Instances
+        - name: Create a simple ec2_instance
+          amazon.aws.ec2_instance:
+            state: present
+            name: "{{ resource_prefix }}-test-basic"
+            instance_type: "{{ ec2_instance_type }}"
+            image_id: "{{ ec2_ami_id }}"        
+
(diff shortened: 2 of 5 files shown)


alinabuzachis (commented):
@chynasan Can you please include integration tests? Thanks.

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,20 @@
 ---
+amazon.aws.ec2_instance_info:
+  query: >-
+    .instances[] | select(. != null) | {
+      name: (.name // .tags.Name // "UNKNOWN"),  # unique identifier for the resource
+      canonical_facts: {
+      id: .instance_id,
+      status: .state.name,
```
alinabuzachis, line 8: Shouldn't it be?
```suggestion
      status: .state,
```

chynasan: Doing it that way returns both the status code integer and the name of the status. I can do it as just .state if you think the code and name are both valuable!

alinabuzachis: Ah gotcha, thanks for explaining. Go with status.name then.

gravesm (commented):
This is missing a changelog.

tests/integration/targets/node_query_instance_ec2/play.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+        - name: List instances by name
+          amazon.aws.ec2_instance_info:
+            filters:
+              "tag:Name": "instance-{{ resource_prefix }}"
+          include_attributes:
+            - kernel
+            - instanceType
+            - userData
+          register: _instances
```
abikouo, line 44:
```suggestion
        - name: List instances by name
          amazon.aws.ec2_instance_info:
            filters:
              "tag:Name": "instance-{{ resource_prefix }}"
            include_attributes:
              - kernel
              - instanceType
              - userData
          register: _instances
```
There is a tabulation issue

tests/integration/targets/node_query_instance_ec2/play.yaml
```diff
@@ -0,0 +1,66 @@
...
+          resource_prefix: "{{ resource_prefix }}"
+    ec2_instance_facts:
+      device_type: "EC2"
+      infra_bucket: "Compute"
+      infra_type: "PublicCloud"
+
+  tasks:
+    - block:
+        # EC2 Instances
+        - name: Create a simple ec2_instance
```
abikouo, line 29: The instance created will exist in the default subnet of the default VPC. It is supposed that the account is configured with both (it is the case of the aws ci account). However, I would rather suggest creating a VPC and a subnet dedicated to the test case, where the EC2 instance will be created.

abikouo (changes requested):

tests/integration/targets/node_query_instance_ec2/play.yaml
chynasan: Thank you! I had pushed this to ask for some help cuz I couldn't get it to run locally and was struggling to figure out why, so this helps!

tests/integration/targets/node_query_instance_ec2/play.yaml
chynasan: Okay, I will add those steps!

abikouo (approved):

tests/integration/targets/node_query_instance_ec2/play.yaml
```diff
@@ -0,0 +1,69 @@
...
+                  tags: "{{ _instances.instances.0.tags }}"
+                facts: "{{ ec2_instance_facts }}"
+
+      always:
+      - name: Delete instance created for tests
+        amazon.aws.ec2_instance:
+          state: absent
+          instance_ids: "{{ _instances.instances | map(attribute='instance_id') | list }}"
+          wait: true
+        ignore_errors: true
```
gravesm: I believe this isn't needed as the `setup_ec2_instance_env` role will delete any instances in the VPC created for this test.

chynasan: ooh, that's rad - I will pull this out, thank you!

gravesm (approved):

beeankha (approved):

alinabuzachis (approved):

PR #2818, opened 2026-01-20
Indirect node count: query for networking load balancer resources

beeankha:
##### SUMMARY

Create a query for indirect node count for Networking - Load Balancers resources (`elb_application_elb_info`, `elb_classic_lb_info`).
Refers to https://issues.redhat.com/browse/ACA-4702

##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/extensions/audit/event_query.yml
+++ b/extensions/audit/event_query.yml
@@ -1,4 +1,20 @@
 ---
+amazon.aws.ec2_instance_info:
+  query: >-
+    .instances[] | select(. != null) | {
+      name: (.name // .tags.Name // "UNKNOWN"),  # unique identifier for the resource
+      canonical_facts: {
+      id: .instance_id,
+      status: .state.name,
+      tags: (.tags // {}),
+      },
+      facts: {
+        infra_type: "PublicCloud",
+        infra_bucket: "Compute",
+        device_type: "EC2"
+        },
+      }
+
 amazon.aws.elb_application_lb_info:
   query: >-
     .load_balancers[] | select(. != null) |
--- a/changelogs/fragments/20260123-indirect-node-count-ec2.yml
+++ b/changelogs/fragments/20260123-indirect-node-count-ec2.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - indirect node count - create query for ec2 (https://github.com/ansible-collections/amazon.aws/pull/2807).
--- a/tests/integration/targets/node_query_instance_ec2/aliases
+++ b/tests/integration/targets/node_query_instance_ec2/aliases
@@ -1,0 +1,2 @@
+cloud/aws
+ec2_instance_info
--- a/tests/integration/targets/node_query_instance_ec2/play.yaml
+++ b/tests/integration/targets/node_query_instance_ec2/play.yaml
@@ -1,0 +1,62 @@
+- hosts: localhost
+  gather_facts: true
+
+  roles:
+    - role: setup_node_query
+    - role: setup_ec2_facts
+    - role: setup_ec2_instance_env
+
+  module_defaults:
+    group/aws:
+      access_key: "{{ aws_access_key }}"
+      secret_key: "{{ aws_secret_key }}"
+      session_token: "{{ security_token | default(omit) }}"
+      region: "{{ aws_region }}"
+
+  vars:
+    ec2_instance_buckets:
+      name: "instance-{{ resource_prefix }}"
+      tags:
+        resource_prefix: "{{ resource_prefix }}"
+    ec2_instance_facts:
+      device_type: "EC2"
+      infra_bucket: "Compute"
+      infra_type: "PublicCloud"
+
+  tasks:
+    - block:
+        # EC2 Instances
+        - name: Create a simple ec2_instance
+          amazon.aws.ec2_instance:
+            state: running
+            name: "{{ ec2_instance_buckets.name }}"
+            instance_type: t3.nano
+            image_id: "{{ ec2_ami_id }}"
+            tags: "{{ ec2_instance_buckets.tags }}"
+            security_group: "{{ security_group_name_1 }}"
+            vpc_subnet_id: "{{ testing_subnet_a.subnet.id }}"
+            wait: false
+          register: result
+
+        - name: List instances by name
+          amazon.aws.ec2_instance_info:
+            instance_ids: "{{ result.instance_ids }}"
+          register: _instances
+
+        - name: Run jq query on result data
+          ansible.builtin.set_fact:
+            node_count: "{{ _instances | to_json | aws_jq(event_node_queries.ec2_instance_info) }}"
+
+        - name: Validate result is as expected
+          ansible.builtin.assert:
+            that:
+              - node_count == expected_result  
+          vars:
+            expected_result: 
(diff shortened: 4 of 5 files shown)


tests/integration/targets/node_query_networking_elb/play.yaml
```diff
@@ -0,0 +1,381 @@
...
+
+  module_defaults:
+    group/aws:
+      access_key: "{{ aws_access_key }}"
+      secret_key: "{{ aws_secret_key }}"
+      session_token: "{{ security_token | default(omit) }}"
+      region: "{{ aws_region }}"
+
+  vars:
+    resource_short: "{{ '%0.8x' % ((16**8) | random(seed=resource_prefix)) }}"
```
alinabuzachis: Why not using `tiny_prefix` instead?

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,51 @@
...
+      facts: {
+        infra_type: "PublicCloud",
+        infra_bucket: "Networking",
+        device_type: "ApplicationLoadBalancer"
+      },
+      extra_fields: {
+      scheme: (.scheme // null),
+      availability_zones: (.availability_zones // []),
+      security_groups: (.security_groups // []),
+      subnets: (.subnets // [])
```
alinabuzachis: I don't think we need to include these extra_fields at this stage. We should prioritize ensuring the basic functionality works first. We can always enrich and extend the feature in future iterations.

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,51 @@
...
+amazon.aws.elb_application_lb_info:
+  query: >-
+    .load_balancers[] | select(. != null) |
+    {
+      canonical_facts: {
+        name: .load_balancer_name,
+        arn: (.load_balancer_arn // null),
+        dns_name: (.dns_name // null),
+        vpc_id: (.vpc_id // null),
+        tags: (.tags // {})
```
alinabuzachis: We should ensure the canonical facts required are met:

```
load_balancer_name -> name,
state -> status,
tags
id (if there's any)
```

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,51 @@
...
+
+amazon.aws.elb_classic_lb_info:
+  query: >-
+    .elbs[] | select(. != null) |
+    {
+      canonical_facts: {
+        name: .load_balancer_name,
+        dns_name: (.dns_name // null),
+        vpc_id: (.vpc_id // null),
+        tags: (.tags // {})
```
alinabuzachis: Same here.

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,51 @@
...
+      facts: {
+        infra_type: "PublicCloud",
+        infra_bucket: "Networking",
+        device_type: "ClassicLoadBalancer"
+      },
+      extra_fields: {
+        scheme: (.scheme // null),
+        availability_zones: (.availability_zones // []),
+        security_groups: (.security_groups // []),
+        subnets: (.subnets // [])
```
alinabuzachis: Same.

tests/integration/targets/node_query_networking_elb/play.yaml
beeankha: The other [`node_query` test uses `resource_prefix`](https://github.com/ansible-collections/amazon.aws/blob/main/tests/integration/targets/node_query_storage_s3/play.yaml#L16-L23) so I was keeping things consistent. 

I've updated this line, are there any other places in this integration test where it would be better for me to switch over to `tiny_prefix` instead?

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,43 @@
...
+    .load_balancers[] | select(. != null) |
+    {
+      canonical_facts: {
+        name: .load_balancer_name,
+        status: (.state.code // null),
+        tags: (.tags // {}),
+        id: (.load_balancer_arn // null),
+        arn: (.load_balancer_arn // null),
+        dns_name: (.dns_name // null),
+        vpc_id: (.vpc_id // null)
```
gravesm: Are these `null` fallbacks necessary? Aren't they redundant?

beeankha: You're correct, these are redundant. Removed!

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,43 @@
...
+    .load_balancers[] | select(. != null) |
+    {
+      canonical_facts: {
+        name: .load_balancer_name,
+        status: .state.code,
+        tags: (.tags // {}),
+        id: .load_balancer_arn,
+        arn: .load_balancer_arn,
+        dns_name: .dns_name,
+        vpc_id: .vpc_id
```
alinabuzachis: I do not think we need these three fields.

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,43 @@
...
+  query: >-
+    .elbs[] | select(. != null) |
+    {
+      canonical_facts: {
+        name: .load_balancer_name,
+        status: "active",
+        tags: (.tags // {}),
+        id: .canonical_hosted_zone_name_id,
+        dns_name: .dns_name,
+        vpc_id: .vpc_id
```
alinabuzachis: Same for these two.

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,43 @@
 ---
+amazon.aws.elb_application_lb_info:
+  query: >-
+    .load_balancers[] | select(. != null) |
+    {
+      canonical_facts: {
+        name: .load_balancer_name,
+        status: .state.code,
+        tags: (.tags // {}),
+        id: .load_balancer_arn,
```
alinabuzachis: If the _info doesn't return an id for the elb, we can remove this field.

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,43 @@
...
+
+amazon.aws.elb_classic_lb_info:
+  query: >-
+    .elbs[] | select(. != null) |
+    {
+      canonical_facts: {
+        name: .load_balancer_name,
+        status: "active",
+        tags: (.tags // {}),
+        id: .canonical_hosted_zone_name_id,
```
alinabuzachis: Same here.

extensions/audit/event_query.yml
```diff
@@ -1,4 +1,36 @@
...
+      },
+    }
+
+amazon.aws.elb_classic_lb_info:
+  query: >-
+    .elbs[] | select(. != null) |
+    {
+      canonical_facts: {
+        name: .load_balancer_name,
+        status: "active",
```
alinabuzachis: Why is this value hardcoded?

beeankha: Unlike application load balancers, the classic load balancer API response doesn't include a `state` or `status` field; I hardcoded it to be consistent with the application load balancer query (i.e. if the classic load balancer exists and is returned by the API, it's effectively "active").

alinabuzachis: If the classic one does not return a status, we should not include this into the canonical facts. We should not hardcode any value in the canonical facts.

beeankha: Removed!

PR #2821, opened 2026-01-23
Document Indirect Node Query Count

alinabuzachis:
##### SUMMARY

Document Indirect Node Query Count

##### ISSUE TYPE

- Docs Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/extensions/audit/README.md
+++ b/extensions/audit/README.md
@@ -90,8 +90,7 @@ Reliability is ensured through integration testing. The **extensions/audit/event
 To validate the node counting logic yourself, you can run any integration test target prefixed with **node_query_**.
 
 #### Prerequisites
-The integration tests rely on the `jq` Python bindings.
-    * Install via pip: `pip install jq`.
+The indirect node counts integration tests have additional dependencies, specifically the `jq` Python bindings. Please run `pip install -r tests/integration/requirements.txt` to ensure all requirements are installed."
 
 #### Command
 **Example: Running the Networking Node Query Test**


extensions/audit/README.md
```diff
@@ -0,0 +1,103 @@
...
+- Why they are required
+- What they enable
+- Their value
+
+## What Are Indirect Node Counts?
+
+Rather than connecting directly to each node to assess its state  of automation(e.g., Service, DB, VM, etc), **indirect node counts** leverage tools like **APIs**, or other management layers to obtain usage information.
+
+In an Ansible role or collection, this might involve:
+- Querying a cloud service for the list of DBs or VMs and their status of if they are being automated
```
gravesm:
```suggestion
- Querying a cloud service for the list of DBs or VMs and their status or if they are being automated
```
Not sure if this is what you meant or not.

extensions/audit/README.md
gravesm, line 1: It's entirely possible I just don't understand how indirect node counting works, but I was under the impression that controller is just using the query files to parse the module return results from a job. This document makes it sound like the counting is an active process that will be making API requests.

alinabuzachis: @gravesm I copied the information from https://github.com/ansible-collections/azure/blob/dev/extensions/audit/README.MD to maintain consistency. I agree with your assessment; that aligns with my understanding as well.

I'm just wondering, if https://github.com/ansible-collections/azure/blob/dev/extensions/audit/README.MD is still accurate https://github.com/ansible-collections/azure/blob/dev/extensions/audit/README.MD or needs more clarification.

@p3ck can you please clarify? @GomathiselviS do you think we need to pull in a specific reviewer for this information?

extensions/audit/README.md
```diff
@@ -0,0 +1,103 @@
+# Indirect Node Counts in Ansible Collections
+
+## Overview
+
+In the context of Ansible automation, **indirect node counts** refer to the practice of calculating or verifying the capacity or availability of computing nodes **through external systems or controllers**, rather than directly querying the nodes themselves. This approach is especially useful in complex, public cloud, or on prem environments .
```
tremble:
```suggestion
In the context of Ansible automation, **indirect node counts** refer to the practice of calculating or verifying the capacity or availability of computing nodes **through external systems or controllers**, rather than directly querying the nodes themselves. This approach is especially useful in complex, public cloud, or on premises environments .
```

My understanding is that it's best to avoid the shortened forms of words like this as it complicates potential translations.

extensions/audit/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+This file explains:
+- What indirect node counts are
+- Why they are required
+- What they enable
+- Their value
+
+## What Are Indirect Node Counts?
+
+Rather than connecting directly to each node to assess its state  of automation(e.g., Service, DB, VM, etc), **indirect node counts** leverage tools like **APIs**, or other management layers to obtain usage information.
```
tremble:
```suggestion
Rather than connecting directly to each node to assess its state of automation (e.g., Service, DB, VM, etc), **indirect node counts** leverage tools like **APIs**, or other management layers to obtain usage information.
```

extensions/audit/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Using indirect node counts via cluster managers:
+- Enables **centralized insight** into resource usage
+- Works well in **managed or disconnected environments**
+- Allows **non-invasive** assessment (e.g., read-only API access)
+
+## What Does It Do?
+
+In practical terms, using indirect node counts in an Ansible collection:
+- Enables **guardrails** to verify into knowing what you are automating
```
tremble:
```suggestion
- Enables **guardrails** to verify what you are automating
```

I'm not 100% sure what the original intent of this sentence was, but "verify into knowing" doesn't quite parse.

extensions/audit/README.md
gravesm: I found https://access.redhat.com/articles/7109910 which seems to confirm how I think this feature works.

extensions/audit/README.md
p3ck: Good catch.  I'll update the Azure README as well.

extensions/audit/README.md
p3ck: As far as I know the Azure README is still accurate.  (granted I didn't write that, it was provided for me).

extensions/audit/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+Reliability is ensured through integration testing. The **extensions/audit/event_query.yml** file is explicitly tested in this collection to verify that the node counting logic works as expected against the supported resources.
+
+### How to Run Integration Tests Locally
+
+To validate the node counting logic yourself, you can run any integration test target prefixed with **node_query_**.
+
+#### Prerequisites
+The integration tests rely on the `jq` Python bindings.
+    * Install via pip: `pip install jq`.
```
abikouo: You also need `botocore` and `boto3` library, this sounds trivial for people familiar with the collection, but it is worth mentioning. We can definitely suggest running `pip install -r tests/integration/requirements.txt` to have the full dependencies installed.

PR #2871, opened 2026-03-02
[ACA-3477] Add SonarQube Cloud Test Coverage Info

beeankha:
##### SUMMARY

Resolves [ACA-3477](https://issues.redhat.com/browse/ACA-3477)

- Implement coverage job in `all_green_check.yml`
- New workflows in `sonarcloud.yml` for downloading coverage artifacts and setting coverage report paths
- `all_green` completes only when linters, sanity, units, and coverage all succeed
- SonarCloud is triggered by that completed run, downloads the `coverage*` artifact(s), and passes them to the scanner so SonarQube Cloud can report unit test coverage
- `pytest-cov` was added to test-`requirements.txt`
- Add SonarQube coverage buttons to `README`

TO DO:
- Add SonarQube info to `CI.md` or a new markdown file (would like CI files reviewed first before I tackle this)

### DIFF
--- a/docs/sonarcloud.md
+++ b/docs/sonarcloud.md
@@ -64,7 +64,7 @@ The CI coverage job uses the environment `ansible2.20-py314-with_constraints` so
 `pyproject.toml` does not define SonarQube-specific settings. It configures tools used by tests and coverage:
 
 - **`[tool.pytest]`**: Pytest options (e.g. `xfail_strict = true`).
-- **`[tool.coverage.report]`**: Coverage report options, including `exclude_lines` for lines that are not required to be covered (e.g. `pragma: no cover`, `raise NotImplementedError`).
+- **`[tool.coverage.report]`**: Coverage report options, including `also_exclude` for lines that are not required to be covered (e.g. `raise NotImplementedError`).
 
 These affect how coverage is generated (and thus what appears in `coverage.xml`), but SonarCloud is configured only via `sonar-project.properties` and the workflow arguments.
 


.github/workflows/sonarcloud.yml
```diff
@@ -60,4 +90,6 @@ jobs:
         env:
           SONAR_TOKEN: ${{ secrets.ANSIBLE_COLLECTIONS_ORG_SONAR_TOKEN_CICD_BOT }}
         with:
-          args: ${{ env.SONAR_ARGS }}
+          args: >-
+            ${{ env.SONAR_ARGS }}
+            ${{ env.COVERAGE_PATHS && format('-Dsonar.python.coverage.reportPaths={0}', env.COVERAGE_PATHS) || '' }}
```
tremble: For consistency, would it make sense to move this up into the "Prepare SonarCloud args" step?

.github/workflows/all_green_check.yml
```diff
@@ -24,17 +24,71 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [units]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
```
john-westcott-iv: You might want to matrix this and the ansible-core versions incase you want/need to test with multiple python/ansible-core versions in the future.

.github/workflows/all_green_check.yml
john-westcott-iv: This is running the scanner on a pull request (so you would get you PRs in Sonar) but not on merge (so your main branch coverage results wouldn't update on a merge). Here is our `on` for gateway unit tests:
```
on:
  pull_request:
  push:    # This is triggered on a PR being merged into devel or a stable branch
    branches:
      - devel
      - 'stable-*'
```

So any pull request activity OR a `push` into devel or one of our stable branches. You likely don't need sanity/linting when something merges into a branch so you might want to consider moving your unit tests out of this file.

.github/workflows/all_green_check.yml
```diff
@@ -24,17 +24,71 @@ jobs:
     uses: ./.github/workflows/sanity.yml # use the callable sanity job to run tests
   units:
     uses: ./.github/workflows/units.yml # use the callable units job to run tests
```
john-westcott-iv: I traced this through, you are calling [your own unit tests](https://github.com/ansible-collections/amazon.aws/blob/main/.github/workflows/units.yml) which has a matrix and calls [an ansible-network workflow which runs pytest](https://github.com/ansible-network/github_actions/blob/main/.github/workflows/unit_source.yml).

This implementation would run a second pytest but only for a specific ansible/python version. My recommendation would be to see if you can request a change to the second workflow to see if they can just spit out coverage files for you. Then pytest is only running in one place. If so, you could move your `Rewrite coverage paths to be repo-relative` step to when you pull the artifacts down.

.github/workflows/sonarcloud.yml
john-westcott-iv, line 26: Same comment on this one. This runs when all_green is completed. Typically unit tests will be the longest running job compared to linting/sanity so thats likely fine. But all_green only runs on a specific pull_request events, not when the pull is merged into its target branch.

.github/workflows/all_green_check.yml
tremble: There's a separate set of unit tests that are running (with a matrix).  The only reason we want this is to be able to grab the coverage report.

Long term, it would be better to grab the coverage reports from those tests, but it's not possible today.

.github/workflows/all_green_check.yml
```diff
@@ -24,17 +24,71 @@ jobs:
...
+  coverage:
+    name: Unit test coverage
```
tremble, line 32: Rather than defining this in yet-another-place.

`tox -e ansible2.20-py314-with_constraints` should already be generating a coverage report (currently in html)

We should be able to expand the tox job to generate an XML report we can upload, and that would likely be faster than trying to get an update into ansible-network.

docs/sonarcloud.md
```diff
@@ -0,0 +1,149 @@
...
+- **Output**: `coverage.xml` is written under the `tox` environment directory (e.g. `.tox/ansible2.20-py314-with_constraints/...`).
+
+The CI coverage job uses the environment `ansible2.20-py314-with_constraints` so that a single, consistent coverage report is produced for SonarCloud.
+
+### `pyproject.toml` (coverage and `pytest`)
+
+`pyproject.toml` does not define SonarQube-specific settings. It configures tools used by tests and coverage:
+
+- **`[tool.pytest]`**: Pytest options (e.g. `xfail_strict = true`).
+- **`[tool.coverage.report]`**: Coverage report options, including `exclude_lines` for lines that are not required to be covered (e.g. `pragma: no cover`, `raise NotImplementedError`).
```
tremble:
```suggestion
- **`[tool.coverage.report]`**: Coverage report options, including `also_exclude` for lines that are not required to be covered (e.g. `raise NotImplementedError`).
```

this was replaced in #2874 so we automatically pull in things like the new exclusion of `if typing.TYPE_CHECKING:`

PR #2889, opened 2026-03-06
Improve unit test performance and move coverage configuration into pyproject.toml

tremble:
##### SUMMARY
Fix unit test performance by providing dummy AWS credentials to boto3.Session in the placebo fixture. When boto3.Session is initialized without credentials, it attempts to connect to the AWS Metadata service, causing delays in unit tests where the fetched credentials won't be used anyway.

Centralizes coverage configuration in pyproject.toml and simplifies tox.ini :
- Enables XML reporting
- Adds missing coverage reporting for code in extensions/ (Event Driven Ansible plugins)
- ignores the generic boto3/botocore ImportError top-level handling (AnsibleAWSModule is responsible for this)
- Sets `include_namespace_packages=true` becuase we don't generally add `__init__.py` (coverage defaults to skipping directories if it's not there "it must be an executable that doesn't need coverage reporting")

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
tests/unit/utils/amazon_placebo_fixtures.py
pyproject.toml
tox.ini

##### ADDITIONAL INFORMATION
Before this change, each unit test using the placebo fixture would experience a delay as boto3 tried to contact the AWS Metadata service. By providing dummy credentials (EXAMPLE_KEY/EXAMPLE_SECRET), we prevent these unnecessary connection attempts during testing.

The coverage configuration consolidation moves settings to pyproject.toml (the standard location) and simplifies the tox.ini command line.

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,8 +62,24 @@ ignore = ["F401", "E402"]
 [tool.pytest]
 xfail_strict = true
 
+[tool.coverage.run]
+source = [
+    "plugins/",
+    "extensions/",
+]
+relative_files = true
+
 [tool.coverage.report]
 exclude_also = [
     # Don't complain if tests don't hit defensive assertion code:
     "raise NotImplementedError",
+    # Handling for botocore not being available is performed by AnsibleAWSModule
+    "try:\n(    import botocore\n|    import boto3\n)*except ImportError:\n    pass",
 ]
+include_namespace_packages = true
+
+[tool.coverage.html]
+directory = "coverage-report"
+
+[tool.coverage.xml]
+output = "coverage.xml"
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -62,8 +62,24 @@ ignore = ["F401", "E402"]
 [tool.pytest]
 xfail_strict = true
 
+[tool.coverage.run]
+source = [
+    "plugins/",
+    "extensions/",
+]
+relative_files = true
```
tremble, line 70: This line negates the need for rewriting the paths, but only works if we set source in pyproject.yaml

pyproject.toml
```diff
@@ -62,8 +62,24 @@ ignore = ["F401", "E402"]
...
+
 [tool.coverage.report]
 exclude_also = [
     # Don't complain if tests don't hit defensive assertion code:
     "raise NotImplementedError",
+    # Handling for botocore not being available is performed by AnsibleAWSModule
+    "try:\n(    import botocore\n|    import boto3\n)*except ImportError:\n    pass",
 ]
+include_namespace_packages = true
```
tremble, line 79: this line negates the need to list the sub-directories (and removes the SonarCloud warnings that `Cannot resolve the file path 'botocore.py' of the coverage report, ambiguity, the file exists in several 'source'.` )

beeankha (approved):

PR #2904, opened 2026-03-13
Lowered the ansible-core minimum version to 2.16.

alinabuzachis:
##### SUMMARY

Changes to requires_ansible are typically considered major, since increasing the minimum supported version can introduce breaking changes for users on older versions of Ansible. In this case, however, the proposal is to lower the requirement to >=2.16, which should not introduce a breaking change. Users already running ansible-core 2.17 or newer would not notice any difference, while it would allow compatibility with environments running 2.16.

Lowered the ansible-core minimum version to 2.16. This expands compatibility and does not change or remove existing functionality.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -58,6 +58,11 @@ jobs:
               "python-version": "3.14"
             },
             {
-              "ansible-version": "stable-2.16"
+              "ansible-version": "stable-2.16",
+              "python-version": "3.13"
+            },
+            {
+              "ansible-version": "stable-2.16",
+              "python-version": "3.14"
             }
           ]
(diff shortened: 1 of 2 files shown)


GomathiselviS (approved):

changelogs/fragments/relax_ansible_core_version.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - meta/runtime.yml - Lowered the ``ansible-core`` minimum version to ``2.16``. This expands compatibility and does not change or remove existing functionality.
```
beeankha, line 2:
```suggestion
  - meta/runtime.yml - Lowered the ``ansible-core`` minimum version to ``2.16``. This expands compatibility and does not change or remove existing functionality (https://github.com/ansible-collections/amazon.aws/pull/2904).
```

PR #2925, opened 2026-03-27
feat: support for feedback in Amazon SQS EDA Plugin

mkanoor:
##### SUMMARY

EDA now supports a feedback mechanism so we can safely delete in flight events from the SQS Queue once we know that the event if needed has been safely stored in a database to survive process outages and restarts.
To safely delete events from the SQS Queue we need to provide a mechanism to indicate that the event if needed has been persisted and wont be lost if the process restarts

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Ansible EDA Source Plugin for SQS

##### ADDITIONAL INFORMATION
2 new optional arguments have been added

* feedback : bool : Default False
* feedback_timeout : int : Default 120 seconds

Also added dev scripts which can be used for testing these locally without having to use EDA Activation or ansible-rulebook.

### DIFF
--- a/extensions/eda/plugins/event_source/aws_sqs_queue.py
+++ b/extensions/eda/plugins/event_source/aws_sqs_queue.py
@@ -73,12 +73,14 @@ options:
         The source plugin should wait for the response to come
         back on this queue before it picks the next event from
         SQS queue.
+    version_added: 11.3.0
   feedback_timeout:
     type: int
     default: 120
     description:
       - Timeout in seconds to wait for feedback from the rule engine
         before raising an exception. Only applies when feedback is enabled.
+    version_added: 11.3.0
   eda_feedback_queue:
     description:
       - Provided automatically by ansible-rulebook when the feedback
@@ -86,6 +88,7 @@ options:
         It allows the system to wait for confirmation that an event has
         been safely persisted in the database before removing it from the
         event queue. Users do not need to provide a value for this manually.
+    version_added: 11.3.0
 """
 
 EXAMPLES = r"""
--- a/extensions/eda/plugins/event_source/aws_sqs_queue.py
+++ b/extensions/eda/plugins/event_source/aws_sqs_queue.py
@@ -1,7 +1,8 @@
 import asyncio
 import json
 import logging
-from typing import Any, TypedDict
+from typing import Any
+from typing import TypedDict
 
 import botocore.exceptions
 from aiobotocore.session import get_session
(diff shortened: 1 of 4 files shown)


brahmanim:
recheck

extensions/eda/plugins/event_source/aws_sqs_queue.py
```diff
@@ -71,109 +96,265 @@
...
         try:
-            response = await client.get_queue_url(**queue_args)
-        except botocore.exceptions.ClientError as err:
-            if (
-                err.response["Error"]["Code"]
-                == "AWS.SimpleQueueService.NonExistentQueue"
-            ):
-                msg = f"Queue {queue_name} does not exist"
-                raise ValueError(msg) from None
+            await asyncio.wait_for(
```
AlexSCorey, line 242: I see here we are awaiting for any message item.  Can we confidently say that we are going to receive feedback in the same order as events are emitted?  If not, we probably need to have something here to keep track of feedback and messages received so that we can be sure we are only deleting messages for events we've already received and processed.

extensions/eda/plugins/event_source/aws_sqs_queue.py
```diff
@@ -71,109 +96,265 @@
...
+
+    meta = {
+        "MessageId": entry["MessageId"],
+        "event": {"uuid": entry["MessageId"]},
+    }
 
-        queue_url = response["QueueUrl"]
+    msg_body = _parse_message_body(entry)
+
+    await queue.put({"body": msg_body, "meta": meta})
```
mkanoor, line 291: @AlexSCorey We are putting 1 event into the queue and we wait for feedback on that event which gets pre-recorded as delete_entry and that is the one that gets deleted from the Amazon SQS Queue.
You can handle sending in multiple events and then wait for the event that came back and use the event.meta.uuid from that to match the SQS Event and then delete it. That has not been implemented in this PR.

extensions/eda/plugins/event_source/aws_sqs_queue.py
```diff
@@ -61,6 +62,30 @@
         messages might be returned). Valid values are V(1) to V(10).
     type: int
     default: 1
+  feedback:
```
alinabuzachis, line 65: Can you please add version_added: 11.3.0 for the new options?

alinabuzachis:
@mkanoor I think https://github.com/ansible-collections/amazon.aws/pull/2954 should fix the devel issues.

alinabuzachis (approved):

brahmanim (approved):

brahmanim:
regate

tremble:
@brahmanim One concern - Looking at the comment from Zuul https://github.com/ansible-collections/amazon.aws/pull/2925#issuecomment-4322207642 the integration tests aren't being triggered.  We probably need to speak to @abikouo and make sure that the test-splitter knows about extensions.

brahmanim:
> @brahmanim One concern - Looking at the comment from Zuul [#2925 (comment)](https://github.com/ansible-collections/amazon.aws/pull/2925#issuecomment-4322207642) the integration tests aren't being triggered. We probably need to speak to @abikouo and make sure that the test-splitter knows about extensions.

The EDA event source plugins integration tests are disabled on purpose.
We are moving to use GHA instead of Zuul, and we have decided not to update the existing Zuul test-splitter with new event source tests.
We will add this ability to the new GitHub action test-splitter, [https://redhat.atlassian.net/browse/ACA-5105](https://redhat.atlassian.net/browse/ACA-5105).

GomathiselviS:
Can we merge this PR? cc: @brahmanim @alinabuzachis

brahmanim:
recheck

brahmanim:
recheck

brahmanim:
recheck

Issue #2851, opened 2026-02-17
Add vpc regional nat gateway


stein321:
### Summary

In November, AWS introduced regional nat gateway https://aws.amazon.com/blogs/networking-and-content-delivery/introducing-amazon-vpc-regional-nat-gateway/. It would be great if this could be added to the existing ec2_vpc_nat_gateway module, or a new module. This would require allowing a list of subnet ids to be passed in.

### Issue Type

Feature Idea

### Component Name

Feature would be added ec2_vpc_nat_gateway module or create a new module named ec2_vpc_regional_nat_gateway.

### Additional Information

I'm using a playbook that creates one nat gateway and one route table per private subnet. This would allow me to use one nat gateway and one route table for all of my private subnets. 

### Code of Conduct

- [x] I agree to follow the Ansible Code of Conduct

PR #2989, opened 2026-06-04
Adding VPC Regional NAT Gateways

chynasan:
##### SUMMARY
Add support for AWS VPC Regional NAT Gateways to the ec2_vpc_nat_gateway module.
Unlike traditional (zonal), NAT gateways that operate in a single AZ and require a subnet, regional NAT gateways are VPC-scoped, automatically span multiple availability zones, don't require a public subnet, and only support public connectivity. See: https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateways-regional.html

This PR adds:

- availability_mode parameter (zonal/regional, default zonal) to control gateway type
- vpc_id parameter, required when availability_mode=regional (replaces subnet_id)
- Validation that connectivity_type=private is rejected for regional gateways
- VPC-based idempotency checks via if_exist_do_not_create for regional gateways
- EIP release skipping for regional gateways (AWS manages EIP lifecycle in automatic mode)
- Improved release_eip handling to release all EIP addresses (not just the first) when deleting multi-address gateways
- Return field documentation for availability_mode, auto_scaling_ips, auto_provision_zones, route_table_id, and per-address availability_zone/availability_zone_id

Manual mode (AvailabilityZoneAddresses, per-AZ EIP configuration) is not presently included as it will introduce complexity which is outside of the scope of this work. 

Addresses https://github.com/ansible-collections/amazon.aws/issues/2851

##### ISSUE TYPE
- Feature Pull Request
- 
##### COMPONENT NAME
ec2_vpc_nat_gateway
ec2_vpc_nat_gateway_info

##### ADDITIONAL INFORMATION
boto3 create_nat_gateway: https://docs.aws.amazon.com/boto3/latest/reference/services/ec2/client/create_nat_gateway.html

Assisted-by: Claude Opus 4.6


changelogs/fragments/ec2_vpc_nat_gateway-regional-mode.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - ec2_vpc_nat_gateway - Added O(availability_mode) and O(vpc_id) parameters for regional NAT gateway support (https://github.com/ansible-collections/amazon.aws/pull/2988).
+  - ec2_vpc_nat_gateway - The O(release_eip) option now correctly releases all EIP addresses when deleting a NAT gateway with multiple addresses (https://github.com/ansible-collections/amazon.aws/pull/2988).
```
tremble: I can't remember if ansidoc is going to correctly format.  This might be better quoted with double-backticks rather than O() and RV()

mjohns91 (commented):
No immediate blockers identified. Do we want to add unit test coverage for the new code? As for the failing `ec2_vpc_nat_gateway : Create new regional NAT gateway` task, I think we need to add permissions to the [`aws-terminator`](https://github.com/mattclay/aws-terminator/blob/main/aws/policy/networking.yaml) networking policies.

chynasan:
recheck

plugins/modules/ec2_vpc_nat_gateway.py
```diff
@@ -328,8 +414,13 @@ def wait_for_status(client, module: AnsibleAWSModule, waiter_name: str, nat_gate
...
 
 def get_nat_gateways(
-    client, subnet_id: Optional[str] = None, nat_gateway_id: Optional[str] = None, states: Optional[List[str]] = None
+    client,
+    subnet_id: Optional[str] = None,
+    nat_gateway_id: Optional[str] = None,
+    states: Optional[List[str]] = None,
+    vpc_id: Optional[str] = None,
 ) -> List[Dict[str, Any]]:
+
```
beeankha:
```suggestion
```

PR #2997, opened 2026-06-11
Updating Link in README from Relative to Absolute

chynasan:
##### SUMMARY
Changed relative link to absolute link in meta/runtime.yml  based on feedback from PE.

This is to address https://redhat.atlassian.net/browse/ACA-6145. I am likely to need an additional PR for adding deprecation for route53.py module state values.

Since this is a docs-only change, no changelog should be needed per: https://docs.ansible.com/projects/ansible/latest/community/development_process.html#community-changelogs

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
README


tremble (commented):
While https://github.com/ansible/ansible/pull/82355 was merged, I don't think it was backported to 2.16.  I don't know which versions of Ansible the EEs are tested against, but if we're still linting against 2.16, the ignore entries probably still need to be there.

As the comments mention, this is technically a bug in ansible-test rather than amazon.aws

mjohns91 (approved):

brahmanim (approved):

hdefazio (approved):

alinabuzachis (approved):

PR #2998, opened 2026-06-12
Add cache format versioning to inventory plugins

tremble:
##### SUMMARY

Add cache format versioning to inventory plugins to automatically invalidate incompatible caches when the cache structure changes. This allows the aws_rds inventory plugin to cache raw RDS instance data instead of formatted inventory, enabling configuration changes (hostvars_prefix/suffix, keyed_groups, compose) to take effect without manually clearing the cache.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
plugin_utils/inventory, aws_rds

##### ADDITIONAL INFORMATION

The cache versioning system works by wrapping cached data with metadata containing a `_cache_format_version` field. Each inventory plugin declares its expected version via the `CACHE_FORMAT_VERSION` class attribute (defaults to 1 for backward compatibility).

When retrieving cached data:
- If the cached version matches the plugin's expected version, data is returned
- If versions mismatch, the cache is invalidated and fresh data is fetched
- Legacy caches without version metadata are assumed to be version 1

The aws_rds plugin now uses version 2, which caches raw RDS instance data rather than pre-formatted inventory. This means users can modify inventory configuration parameters and see the changes applied on the next run without needing to manually clear the cache.

Assisted-by: Claude Sonnet 4.5

### DIFF
--- a/changelogs/fragments/inventory-cache-versioning.yml
+++ b/changelogs/fragments/inventory-cache-versioning.yml
@@ -1,4 +1,5 @@
 minor_changes:
   - plugin_utils/inventory - add cache format versioning to automatically invalidate incompatible caches when inventory plugins change cache structure (https://github.com/ansible-collections/amazon.aws/pull/XXXX).
+  - aws_rds - cache raw RDS instance data instead of formatted inventory to allow configuration changes (hostvars_prefix/suffix, keyed_groups, compose) to take effect without clearing the cache (https://github.com/ansible-collections/amazon.aws/pull/XXXX).
 trivial:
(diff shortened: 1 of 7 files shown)


tremble:
@Romi1495 an alternative approach to #2996

Over in aws_ec2 we didn't hit this issue because instead of caching the processed data, we cache the "raw" data.  By taking the same approach in the RDS modules we can avoid duplicating the logic.

Romi1495:
@tremble looks good to me! Thanks for finding this

PR #3034, opened 2026-07-16
fix: add UUID validation to aws_sqs_queue EDA source plugin

ttuffin:
The plugin was writing to meta["event"]["uuid"], which is a nested sub-dict that ansible-rulebook's insert_meta_info filter never reads, causing every event to receive a random UUID4 and eda-server validation failures when the SQS MessageId was not a valid RFC 4122 UUID.

Fixes:
- Set meta["uuid"] (flat) to match the schema expected by ansible-rulebook and eda-server
- Validate the SQS MessageId before use; fall back to a deterministic UUID5 when invalid (AWS guarantees MessageId is always a valid UUID in practice, so the fallback is defensive)
- Preserve the original MessageId under the existing meta["MessageId"] key; no additional key is added since the value is already present there
- Emit a one-time warning when an invalid MessageId triggers the fallback

Refs: AAP-77079

Assisted-by: Claude

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
EDA

### DIFF
--- a/changelogs/fragments/3034-aws_sqs_queue-uuid-validation.yml
+++ b/changelogs/fragments/3034-aws_sqs_queue-uuid-validation.yml
@@ -1,0 +1,4 @@
+bugfixes:
+  - aws_sqs_queue - Fix UUID generation in EDA event source plugin to set ``meta.uuid`` as a flat key readable by ansible-rulebook, validate the SQS ``MessageId`` as a proper RFC 4122 UUID, and fall back to a deterministic UUID5 when invalid (https://github.com/ansible-collections/amazon.aws/pull/3034).
+trivial:
+  - aws_sqs_queue - Add unit tests for UUID validation, valid UUID passthrough, and invalid UUID fallback behaviour in the EDA event source plugin (https://github.com/ansible-collections/amazon.aws/pull/3034).
--- a/extensions/eda/plugins/event_source/aws_sqs_queue.py
+++ b/extensions/eda/plugins/event_source/aws_sqs_queue.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 import logging
+import uuid
 from typing import Any
 from typing import TypedDict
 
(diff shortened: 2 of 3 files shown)


ttuffin:
Hi @alinabuzachis @GomathiselviS, I do not have permission to add reviewers. Can you please take a look at this? It's pending review from EDA team as well. Once we have that we should be good to merge.

kaiokmo (approved):

brahmanim (changes requested):
Please add a changelog file under https://github.com/ansible-collections/amazon.aws/tree/main/changelogs/fragments.
Add `aws_sqs_queue` plugin changes under `bugfixes`.
Add unit tests under `trivial`.

ttuffin:
@brahmanim I have added the changelog fragment as requested.

brahmanim (approved):

changelogs/fragments/3034-aws_sqs_queue-uuid-validation.yml
```diff
@@ -0,0 +1,4 @@
+bugfixes:
+  - aws_sqs_queue - Fix UUID generation in EDA event source plugin to set ``meta.uuid`` as a flat key readable by ansible-rulebook, validate the SQS ``MessageId`` as a proper RFC 4122 UUID, and fall back to a deterministic UUID5 when invalid (https://github.com/ansible-collections/amazon.aws/pull/3034).
+trivial:
```
abikouo, line 3: Not really necessary to add this since you already added a `bugfixes` section

abikouo (approved):

beeankha (approved):

oraNod (approved):
Thanks @ttuffin

changelogs/fragments/3034-aws_sqs_queue-uuid-validation.yml
ttuffin: @abikouo I add it at @brahmanim's request. Let me know if you want it removed.

brahmanim:
Hi @ttuffin, would you mind rebasing your PR?
We made some changes to our GitHub Actions workflows to support merging your PR.

brahmanim:
Hi @ttuffin, can you please manually backport to stable-10? See instructions in the above patchback.
Also, you can take #3059 as a reference.

ttuffin:
Hi @brahmanim, this source plugin doesn't exist in stable-10, so we can disregard that backport. Thanks for merging!

PR #3035, opened 2026-07-16
docs: add local unit test and linting instructions to CONTRIBUTING.md

ttuffin:
I realised there was a lack of clear documentation for running unit tests and linters while developing the fix for https://github.com/ansible-collections/amazon.aws/pull/3034. The existing docs only covered integration tests. Unit tests require tox rather than raw pytest or ansible-test due to the ansible_collections namespace resolution needed for fully-qualified imports. Linting requires black-lint and flake8-lint tox environments.

##### ISSUE TYPE
- Docs Pull Request


ttuffin:
Hi @alinabuzachis @GomathiselviS, I do not have permission to add reviewers. Can you please take a look at this?

CONTRIBUTING.md
```diff
@@ -100,6 +100,64 @@ Then create a pull request.
...
+First, install `tox` into your virtual environment:
+
+```bash
+pip install tox
+```
+
+To run all unit tests:
+
+```bash
+tox -e ansible2.20-py312-without_constraints
```
tremble, line 121: alternatively

`tox -m unit-oldest` or `tox -m unit-newest`

CONTRIBUTING.md
```diff
@@ -100,6 +100,64 @@ Then create a pull request.
...
+**Flake8** (style and error checking):
+
+```bash
+tox -e flake8-lint -- <path/to/file>
+```
+
+To auto-fix formatting issues (rather than just checking), use the `black` environment:
+
+```bash
+tox -e black -- <path/to/file>
```
tremble: Even simpler:

```suggestion
tox -m format
```

tremble (commented):

There are actually more than just the two linters in play.  This was already documented, however it's not necessarily where folks are looking for it (as evidenced by you not finding it):

https://github.com/ansible-collections/amazon.aws/blob/main/docs/docsite/rst/dev_guidelines.rst#code-formatting

ttuffin:
@tremble thanks for the review. Based on your comment about this already being covered in dev_guidances, do you still see value in adding this information to the contributing guide? Or shall I close the PR?

mjohns91 (commented):
I don't think adding the details here for local testing is a problem. Seems reasonable enough.

CONTRIBUTING.md
```diff
@@ -100,6 +100,64 @@ Then create a pull request.
...
+See `tox.ini` for the full list of available environments (ansible-core versions and
+Python version combinations).
+
+> **Note:** Do not use `pytest` or `ansible-test units` directly — they will fail with
+> `ModuleNotFoundError: No module named 'ansible_collections'` because the collection
+> namespace is not on `sys.path` without the tox setup.
+
+### Linting
+
+Two linters are enforced in CI. Run them against changed files before submitting a PR:
```
oraNod, line 141: I believe that there are other linters that run in CI. Looking at: https://github.com/ansible-collections/amazon.aws/blob/14294b0f70cb0a8c05cf8f71e4b407125acfe910/.github/workflows/linters.yml#L11

That should run all the linters with `labels = lint` in the tox config such as https://github.com/ansible-collections/amazon.aws/blob/14294b0f70cb0a8c05cf8f71e4b407125acfe910/tox.ini#L129

So this section could be reduced down to `tox -m lint` instead of individual commands.

CONTRIBUTING.md
```diff
@@ -100,6 +100,64 @@ Then create a pull request.
...
+tox -e black-lint -- <path/to/file>
+```
+
+**Flake8** (style and error checking):
+
+```bash
+tox -e flake8-lint -- <path/to/file>
+```
+
+To auto-fix formatting issues (rather than just checking), use the `black` environment:
```
oraNod, line 155:
```suggestion
To automatically fix any formatting issues, use the following command:
```

CONTRIBUTING.md
```diff
@@ -100,6 +100,64 @@ Then create a pull request.
...
+
+```bash
+pip install tox
+```
+
+To run all unit tests:
+
+```bash
+tox -e ansible2.20-py312-without_constraints
+```
```
oraNod, line 122: I feel like this is somewhat misleading. I ran into this myself recently while looking at a pull request in this collection. There are multiple combinations of `ansible-core` and Python versions in the `envlist` and CI runs the lot.

Reading "all unit tests" might cause contributors to think they've fully covered their changes and not just one combo. So I'd rephrase to say something like:

"To run a specific environment that matches your Python version:" or something like that.

For all unit tests, I believe you would need `tox -m unit`.

CONTRIBUTING.md
```diff
@@ -100,6 +100,64 @@ Then create a pull request.
...
+To run a specific test file or filter by test name, use the `-k` flag rather than
+passing a file path directly (file paths resolve relative to the repo root and break
+the collection namespace resolution inside tox):
+
+```bash
+tox -e ansible2.20-py312-without_constraints -- -k test_aws_sqs_queue -v
+```
+
+See `tox.ini` for the full list of available environments (ansible-core versions and
+Python version combinations).
```
oraNod, line 133: What about recommending `tox list` to discover available environments? That might be more straightforward for folks than digging through `tox.ini`.

CONTRIBUTING.md
```diff
@@ -100,6 +100,64 @@ Then create a pull request.
...
+
+> **Note:** Do not use `pytest` or `ansible-test units` directly — they will fail with
+> `ModuleNotFoundError: No module named 'ansible_collections'` because the collection
+> namespace is not on `sys.path` without the tox setup.
+
```
oraNod, line 138: I feel like `tox -e diff-cover` is worth a mention in here. https://github.com/ansible-collections/amazon.aws/blob/14294b0f70cb0a8c05cf8f71e4b407125acfe910/tox.ini#L89

PR #3038, opened 2026-07-21
fix(ec2_metadata_facts): improve support for tags with dashes

RBastianini:
##### SUMMARY
Fixed a bug that prevented from adding tags whose name contains dashes or colons to the "instance_tags" dictionary.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ec2_metadata_facts

##### ADDITIONAL INFORMATION
ec2 tags containing `:` and `-` were not added to the `instance_tags` dictionary, but their name would still appear in the `instance_tags_keys` one.
This was caused by the following order of operations:
1) Copying the tag names into the `instance_tags_keys` dictionary,
2) Calling `Ec2Metadata.fix_invalid_varnames`, on the `data` dictionary, renaming the dictionary keys,
3) Using `instance_tags_keys` to fetch corresponding values from the `data` dictionary.

Since 2) would rename dictionary entries with `:` and `-`, looking their value up using the original name from `instance_tags_keys` would always fail.


changelogs/fragments/parse-ec2-tags-with-dashes.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+  - ec2_metadata_facts: Fix parsing of tags with dashes in the name. Before such tags would be reported in the "instance_tags_keys" dictionary, but they would be missing from the "instance_tags" one.
```
oraNod:
```suggestion
  - ec2_metadata_facts: Fix parsing of tags with dashes in the name. Before such tags would be reported in the "instance_tags_keys" dictionary, but they would be missing from the "instance_tags" one.

```

plugins/modules/ec2_metadata_facts.py
```diff
@@ -684,6 +683,7 @@ def run(self):
         # Maintain old key for backwards compatibility
         if "ansible_ec2_instance_identity_document_region" in data:
             data["ansible_ec2_placement_region"] = data["ansible_ec2_instance_identity_document_region"]
+        data = self.fix_invalid_varnames(data)
```
oraNod, line 688: @RBastianini Hey, thanks for the contribution. However I believe this change results in `ansible_ec2_instance_tags` being emptied rather than `kebab-case-key` tags being dropped.

Here is what I found:

`_mangle_fields()` stores tag values under keys like `ansible_ec2_tags-instance-<name>` (dash-joined). 

`get_instance_tags()` looks them up with a hardcoded underscore string: `data.get("ansible_ec2_tags_instance_{}".format(key))`.

The reason this matched is that `fix_invalid_varnames()` ran before the lookup and converted the dashes to underscores. Moving it to run after the lookup means the lookup never matches anything.

A possible solution is to modify `get_instance_tags()` to derive the lookup key in the same way that `_mangle_fields` does instead of the hardcoded underscore-joined literal. For example:

```
def get_instance_tags(self, tag_keys, data):
      tags = {}
      tag_path = self.uri_instance_tags[len(self.uri_meta):].strip("/")  # "tags/instance"
      for key in tag_keys:
          mangled_key = self._prefix % "-".join(tag_path.split("/") + [key])
          value = data.get(mangled_key)
          if value is not None:
              tags[key] = value
      return tags
```

It would also be ideal to have a unit test for `get_instance_tags` with a mocked dict to assert that tags end up in `instance_tags` correctly.

RBastianini: Thanks for your review and for the detailed explanation. I have included your suggestions and added a unit test.

plugins/modules/ec2_metadata_facts.py
```diff
@@ -684,6 +685,7 @@ def run(self):
...
```
oraNod, line 688: Looks like the backwards compatibility portion needs to come before `data = self.fix_invalid_varnames(data)`.

```
data = self.fix_invalid_varnames(data)
  if "ansible_ec2_instance_identity_document_region" in data:
      data["ansible_ec2_placement_region"] =
  data["ansible_ec2_instance_identity_document_region"]
  return data
```

Otherwise `fix_invalid_varnames` converts to `ansible_ec2_instance_identity_document_region` after the backwards compatibility check and `ansible_ec2_placement_region` does not get set.

You can see this by adding the following to `tests/unit/plugins/modules/test_ec2_metadata_facts.py`:

```
def test_placement_region_backwards_compat(ec2_instance):
    # Simulate what _mangle_fields produces before fix_invalid_varnames
    data = {
        "ansible_ec2_instance-identity-document-region": "us-east-1",
    }

    if "ansible_ec2_instance_identity_document_region" in data:
        data["ansible_ec2_placement_region"] = data["ansible_ec2_instance_identity_document_region"]

    data = ec2_instance.fix_invalid_varnames(data)

    assert data.get("ansible_ec2_placement_region") == "us-east-1"
```

Then run the tests with something like (in my case I'm using Python3.13): 

```
tox -e ansible2.19-py313-without_constraints -- tests/unit/plugins/modules/test_ec2_metadata_facts.py::test_placement_region_backwards_compat -v
```

tests/unit/plugins/modules/test_ec2_metadata_facts.py
```diff
@@ -99,3 +99,21 @@ def test__fetch_user_data_plain(m_fetch_url, ec2_instance):
...
+        "ansible_ec2_tags-instance-CamelCaseName": "CamelCaseValue",
+        "ansible_ec2_tags-instance-snake_case_key": "snake_case_value",
+        "ansible_ec2_tags-instance-kebab-case-name": "kebab-case-value",
+    }
+    expected = {
+        "CamelCaseName": "CamelCaseValue",
+        "snake_case_key": "snake_case_value",
+        "kebab-case-name": "kebab-case-value",
+    }
+    assert ec2_instance.get_instance_tags(instance_tags_keys, data) == expected
```
oraNod, line 119: Nitpick. Needs a trailing new line at EOF.

plugins/modules/ec2_metadata_facts.py
```diff
@@ -655,10 +655,12 @@ def fetch_session_token(self, uri_token):
 
     def get_instance_tags(self, tag_keys, data):
         tags = {}
+        tag_path = self.uri_instance_tags[len(self.uri_meta):].strip("/") # "tags/instance"
         for key in tag_keys:
-            value = data.get("ansible_ec2_tags_instance_{}".format(key))
-            if value is not None:
-                tags[key] = value
+          mangled_key = self._prefix % "-".join(tag_path.split('/') + [key])
```
oraNod, line 660: Hmm. I guess this one is on me but this doesn't seem very DRY looking at https://github.com/ansible-collections/amazon.aws/blob/a07e0ff8e564b488b0278822e2fdd21f679e6464/plugins/modules/ec2_metadata_facts.py#L572

Rather than rederiving how `mangle_fields` builds keys, maybe you could use `_fetch` instead:

```
def get_instance_tags(self, tag_keys):
      tags = {}
      for key in tag_keys:
          value = self._fetch(f"{self.uri_instance_tags}/{key}")
          if value is not None:
              tags[key] = value
      return tags
```

tests/unit/plugins/modules/test_ec2_metadata_facts.py
```diff
@@ -99,3 +99,21 @@ def test__fetch_user_data_plain(m_fetch_url, ec2_instance):
...
+    instance_tags_keys = [
+        "CamelCaseName",
+        "snake_case_key",
+        "kebab-case-name",
+    ]
+    data = {
+        "ansible_ec2_tags-instance-CamelCaseName": "CamelCaseValue",
+        "ansible_ec2_tags-instance-snake_case_key": "snake_case_value",
+        "ansible_ec2_tags-instance-kebab-case-name": "kebab-case-value",
+    }
```
oraNod, line 113:
```suggestion
    responses = {
        "http://169.254.169.254/latest/meta-data/tags/instance/CamelCaseName": "CamelCaseValue",
        "http://169.254.169.254/latest/meta-data/tags/instance/snake_case_key": "snake_case_value",
        "http://169.254.169.254/latest/meta-data/tags/instance/kebab-case-name": "kebab-case-value",
    }
    ec2_instance._fetch = lambda uri: responses.get(uri)
```

oraNod: If you go with `_fetch` as in my previous comment, you'll need to update the unit test as well.

oraNod (commented):
Hey @RBastianini some more review comments for you. Thanks again for the contribution. I hope you don't mind me adding a bit of churn with the review. To be honest, I'm learning about this collection myself and finding it helpful.

RBastianini:
I've been a bit busy with life, but I did not forget about this PR. I hope to pick it back up in the next days.

PR #3039, opened 2026-07-23
Replace ansible.module_utils.six with Python stdlib equivalents

ChihweiLHBird:
``ansible.module_utils.six`` is deprecated in ansible-core 2.21 and is scheduled for removal in 2.24. Every use of it in this collection has a direct Python 3 standard library equivalent, so this replaces them:

- ``urllib.parse``
- ``urllib.error``
- ``urllib``

No behaviour change on Python 3.

This intentionally drops Python 2 fallbacks from code that runs on the managed node. Python 2.7 targets were last supported by ansible-core 2.16 (EOL).

Assisted-by: Grok 4.5 (xAI)


oraNod (approved):
LGTM. Thanks @ChihweiLHBird

tremble (approved):
ec2_metadata_facts is the only module that is **explicitly** intended to run as a managed node.  It deliberately doesn't need botocore, which means that our general Python requirement doesn't apply.

As mentioned in the PR, with Ansible 2.16 support for RHEL7 as a target OS was dropped.  We currently state:
> Tested with Ansible Core 2.17.0 and later, and the current development version of Ansible. Ansible Core versions prior to 2.17.0 are not supported.

This change was part of Release 11.0.0

While I'm +1 on merging this change, I'm torn as to if we need to explicitly call out the change to ec2_metadata_facts as a breaking change.  We're taking the deliberate step which we *know* will break things if someone's using this to manage an RHEL7 box.  However, runtime.yml does already say we need 2.17, which in turn will already have broken things for folks connecting to RHEL7.  

See also: https://github.com/ansible-collections/amazon.aws/issues/1358

ChihweiLHBird:
Hi @tremble @oraNod, thanks for the review! If the change in ec2_metadata_facts might be a breaking change, should I update the changelog for it? Or any other action required from my side?

PR #3033, opened 2026-07-15
refactor(rds): decompose module_utils/rds.py into _rds/ private subpackage

mjohns91:
##### SUMMARY
Refactor `module_utils/rds.py` into a `_rds/` private subpackage following the established `_s3/`, `_autoscaling/`, `_elbv2/` pattern used in this collection.

- Decompose the monolithic `rds.py` (790 lines) into focused submodules: `common.py` (error classes, method routing), `waiters.py`, `transformations.py`, `api.py`, and `tags.py`
- Rewrite `rds.py` as a thin facade re-exporting all public symbols — all 13 consuming modules continue importing from `rds.py` unchanged
- Remove redundant error handling and manual polling from `rds_instance.py` in favor of shared functions from #2119
- Reorganize `test_rds.py` into a `rds/` subdirectory with per-submodule test files matching the `s3/` test pattern
- Add comprehensive unit test coverage targeting 80% on all new submodules: waiters, api, transformations, common, and tags

##### ISSUE TYPE
Refactor Pull Request

##### COMPONENT NAME
rds (module_utils), rds_instance

##### ADDITIONAL INFORMATION
Continuation of #2119, which migrated `rds_instance` and `rds_instance_info` to use shared functions from `module_utils/rds.py`. This PR completes the refactor by cleaning up residual legacy patterns and decomposing the monolithic utility module into maintainable submodules.

Review feedback addressed:
- Removed dead `db_cluster_promoting` override that could never match
- Added explicit `return True` after `fail_json_aws` calls for type-safety
- Added non-`ClientError` guard in `handle_errors` to prevent `exception.response` access on `BotoCoreError`
- Fixed `update_iam_roles` to accumulate `changed` state with `|=` instead of overwriting each iteration
- Strengthened `test_get_snapshot_success` to set only the correct describe mock and assert the other was not called
- Added tests for changed accumulation and empty role lists in `update_iam_roles`

All unit tests pass (127 module_utils tests, 14 rds_instance tests, 141 total). Sanity tests clean.

Assisted-by: Claude Opus 4.6

PR #3040, opened 2026-07-24
Refactor for rds_cluster_info

chynasan:
##### SUMMARY

This work is largely based on the work done in https://github.com/ansible-collections/amazon.aws/pull/3033 and https://github.com/ansible-collections/amazon.aws/pull/2782. It was written to address ACA-2472.

Adds centralized describe_db_clusters() to module_utils/rds.py with @RDSErrorHandler and @AWSRetry decorators, matching  the existing pattern for describe_db_instances, describe_db_snapshots, and describe_db_cluster_snapshots:

- Refactor rds_cluster_info to use the shared describe_db_clusters and AnsibleRDSError, matching the rds_instance_info pattern
- Add DBClusterNotFoundFault to RDSErrorHandler._is_missing() so the error handler covers the actual AWS error code used by DescribeDBClusters
- Add defensive return True after fail_json_aws in handle_errors for the non-ClientError path (from PR #3033 review feedback)

What was removed from rds_cluster_info

- Local _describe_db_clusters function (duplicated across 3 files)
- try: import botocore and raw botocore.exceptions handling
- is_boto3_error_code import
- Double retry wrapping (was on both the client and the local function)
- Unnecessary try/except around module.client("rds")

What was added

- Type hints and docstring on cluster_info()
- AnsibleRDSError error handling in main() (matching rds_instance_info)
- Unit tests for rds_cluster_info (5 tests) and describe_db_clusters (3 tests)

Co-Authored-By: Claude Opus 4.6

##### ISSUE TYPE
- Refactoring Pull Request

##### COMPONENT NAME
rds_cluster_info

##### ADDITIONAL INFORMATION

Assisted-by: Claude Opus 4.6

PR #3041, opened 2026-07-24
Refactor rds_cluster module

chynasan:
##### SUMMARY
This work is largely based on the work done in https://github.com/ansible-collections/amazon.aws/pull/3033 and https://github.com/ansible-collections/amazon.aws/pull/2782. It was written to address ACA-2472.

- Remove global module and global client from main(), pass as explicit parameters to all 8 functions that used them
  implicitly
- Replace eval() dispatch (3 locations) and 7 manual parameter-filtering functions (get_create_options, get_modify_options, get_delete_options, get_restore_s3_options, get_restore_snapshot_options, get_restore_cluster_options, get_backtrack_options) with format_rds_client_method_parameters() from  module_utils/rds.py
- Replace local _describe_db_clusters with centralized describe_db_clusters from module_utils/rds.py (added in PR 1)
- Replace raw botocore.exceptions handling with properly imported BotoCoreError/ClientError and AnsibleRDSError
- Add type hints and docstrings to all functions

#### Bug fixes
- add_role / backtrack_cluster signature mismatch: Both were defined with 1 parameter (params) but called with 3 (client, module, params). These code paths would crash with TypeError at runtime. Fixed signatures to accept all 3 parameters.
- CloudWatch logs modification silently ignored: get_modify_options listed CloudwatchLogsExportConfiguration (the boto3 API parameter name) but the parameters dict uses EnableCloudwatchLogsExports (from arg_spec_to_rds_params conversion). Since the keys didn't match, cloudwatch logs changes were silently filtered out before reaching changing_cluster_options. Fixed by extracting EnableCloudwatchLogsExports before parameter filtering and re-adding it for changing_cluster_options to process.
- Manual parameter lists (which had accumulated copy-paste artifacts like duplicate entries) replaced with dynamic format_rds_client_method_parameters().

Net result
- 161 fewer lines of code (1436 → 1275)
- No eval(), no global, no duplicated parameter lists
- All function signatures are explicit about their dependencies

Test plan
- [ ] ansible-test units tests/unit/plugins/modules/test_rds_cluster.py tests/unit/plugins/module_utils/test_rds.py -v
- [ ] Existing integration tests for rds_cluster_create, rds_cluster_modify, rds_cluster_states, rds_cluster_snapshot (these exercise rds_cluster and rds_cluster_info as a dependency)
- [ ] Verify no changes to module parameters or return values (internal refactor only)

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module/Plugin Pull Request
- Refactoring Pull Request
- Test Pull Request

##### COMPONENT NAME
rds_cluster

##### ADDITIONAL INFORMATION

I am still running all the integration tests here, so this may be updated further before I request any specific reviews.

Assisted-by: Claude Opus 4.6

### DIFF
--- a/plugins/modules/rds_cluster_info.py
+++ b/plugins/modules/rds_cluster_info.py
@@ -245,48 +245,47 @@ clusters:
 """
 
 
-try:
-    import botocore
-except ImportError:
-    pass  # handled by AnsibleAWSModule
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Union
 
 from ansible.module_utils.common.dict_transformations import camel_dict_to_snake_dict
 
-from ansible_collections.amazon.aws.plugins.module_utils.botocore import is_boto3_error_code
 from ansible_collections.amazon.aws.plugins.module_utils.modules import AnsibleAWSModule
+from ansible_collections.amazon.aws.plugins.module_utils.rds import AnsibleRDSError
+from ansible_collections.amazon.aws.plugins.module_utils.rds import describe_db_clusters
 from ansible_collections.amazon.aws.plugins.module_utils.rds import get_tags
-from ansible_collections.amazon.aws.plugins.module_utils.retries import AWSRetry
 from ansible_collections.amazon.aws.plugins.module_utils.transformation import ansible_dict_to_boto3_filter_list
 
 
-@AWSRetry.jittered_backoff(retries=10)
-def _describe_db_clusters(client, **params):
-    try:
-        paginator = client.get_paginator("describe_db_clusters")
-        return paginator.paginate(**params).build_full_result()["DBClusters"]
-    except is_boto3_error_code("DBClusterNotFoundFault"):
-        return []
+def cluster_info(
+    client, module: AnsibleAWSModule, cluster_id: Optional[str], filters: Optional[Dict[str, Union[str, List]]]
+) -> List[Dict[str, Any]]:
+    """Returns attributes of DB cluster(s), optionally filtered by ID and filters.
 
+    Parameters:
+        client: boto3 rds client
+        module: AnsibleAWSModule
+        cluster_id: Unique identifier of DB cluster to describe
+        filters: Additional boto3-supported filters
 
-def cluster_info(client, module):
-    cluster_id = module.params.get("db_cluster_identifier")
-    filters = module.params.get("filters")
-
-    params = dict()
+    Returns:
+        List of cluster attribute dicts in snake_case format
+    """
+    params = {}
     if cluster_id:
         params["DBClusterIdentifier"] = cluster_id
     if filters:
         params["Filters"] = ansible_dict_to_boto3_filter_list(filters)
 
-    try:
-        result = _describe_db_clusters(client, **params)
-    except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
-        module.fail_json_aws(e, "Couldn't get RDS cluster information.")
+    results = describe_db_clusters(client, **params)
 
-    for cluster in result:
+    for cluster in results:
         cluster["Tags"] = get_tags(client, module, cluster["DBClusterArn"])
 
-    return dict(changed=False, clusters=[camel_dict_to_snake_dict(cluster, ignore_list=["Tags"]) for cluster in result])
+    return [camel_dict_to_snake_dict(cluster, ignore_list=["Tags"]) for cluster in results]
 
 
 def main():
@@ -300,12 +299,15 @@ def main():
         supports_check_mode=True,
     )
 
-    try:
-        client = module.client("rds", retry_decorator=AWSRetry.jittered_backoff(retries=10))
-    except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
-        module.fail_json_aws(e, msg="Failed to connect to AWS.")
+    client = module.client("rds")
 
-    module.exit_json(**cluster_info(client, module))
+    cluster_id = module.params.get("db_cluster_identifier")
+    filters = module.params.get("filters")
+
+    try:
+        module.exit_json(changed=False, clusters=cluster_info(client, module, cluster_id, filters))
+    except AnsibleRDSError as e:
+        module.fail_json_aws(e)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


plugins/modules/rds_cluster_info.py
```diff
@@ -300,12 +299,15 @@ def main():
         supports_check_mode=True,
     )
 
-    try:
-        client = module.client("rds", retry_decorator=AWSRetry.jittered_backoff(retries=10))
-    except (botocore.exceptions.ClientError, botocore.exceptions.BotoCoreError) as e:
-        module.fail_json_aws(e, msg="Failed to connect to AWS.")
+    client = module.client("rds")
```
abikouo, line 302: The old code created the client with retry_decorator=AWSRetry.jittered_backoff(retries=10). The new code uses module.client("rds") with no retry decorator. The centralized describe_db_clusters has its own @AWSRetry.jittered_backoff()
  decorator, but any other client calls (like get_tags which calls list_tags_for_resource) will lose retry behavior. This is a subtle behavioral change that could cause flaky failures in production.

abikouo (commented):
A little remark regarding the client initialization

plugins/modules/rds_cluster.py
```diff
@@ -767,42 +767,46 @@
       sample: "sg-12345678"
 """
 
+
+from typing import Any
+from typing import Dict
+from typing import Optional
+from typing import Tuple
```
oraNod: I believe this import is not used anywhere and can be dropped.

plugins/modules/rds_cluster.py
```diff
@@ -1160,24 +993,34 @@ def changing_cluster_options(modify_params, current_cluster):
...
+) -> bool:
+    """Create a new cluster or modify an existing one."""
     changed = False
 
     if not cluster:
-        if parameters.get("Tags") is not None:
-            parameters["Tags"] = ansible_dict_to_boto3_tag_list(parameters["Tags"])
-
-        call_method(client, module, method_name, eval(method_options_name)(parameters))
+        filtered = format_rds_client_method_parameters(client, module, parameters, method_name, format_tags=True)
```
oraNod, line 1001: With this change I think we can drop `from ansible_collections.amazon.aws.plugins.module_utils.tagging import ansible_dict_to_boto3_tag_list` on line 794.

changelogs/fragments/rds_cluster-refactor.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - rds_cluster - Removed module-level globals, replaced ``eval()`` dispatch with ``format_rds_client_method_parameters()``, and use shared ``describe_db_clusters`` from ``module_utils/rds.py`` for consistent error handling (https://github.com/ansible-collections/amazon.aws/pull/XXXX).
+bugfixes:
+  - rds_cluster - Fixed ``add_role`` and ``backtrack_cluster`` function signature mismatch that would cause a TypeError when adding IAM roles or backtracking clusters (https://github.com/ansible-collections/amazon.aws/pull/XXXX).
```
oraNod: Might be an obvious one but we should be sure to update `pull/XXXX` to `pull/3041` before merging.

oraNod (commented):
Thanks @chynasan

changelogs/fragments/rds_cluster-refactor.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - rds_cluster - Removed module-level globals, replaced ``eval()`` dispatch with ``format_rds_client_method_parameters()``, and use shared ``describe_db_clusters`` from ``module_utils/rds.py`` for consistent error handling (https://github.com/ansible-collections/amazon.aws/pull/XXXX).
```
oraNod:
```suggestion
  - rds_cluster - Removed module-level globals, replaced ``eval()`` dispatch with ``format_rds_client_method_parameters()``, and use shared ``describe_db_clusters`` from ``module_utils/rds.py`` for consistent error handling (https://github.com/ansible-collections/amazon.aws/pull/3041).
```

changelogs/fragments/rds_cluster-refactor.yml
oraNod:
```suggestion
  - rds_cluster - Fixed ``add_role`` and ``backtrack_cluster`` function signature mismatch that would cause a TypeError when adding IAM roles or backtracking clusters (https://github.com/ansible-collections/amazon.aws/pull/3041).
```

oraNod (commented):
Probably more helpful to make a suggestion... 😺

mjohns91 (commented):
I think this looks good. Happy to approve after other suggestions are implemented!

plugins/modules/rds_cluster.py
chynasan: Good catch, I removed this. Thank you!

plugins/modules/rds_cluster.py
chynasan: We are using Optional, Any and Dict here I believe, but I did go ahead and remove Tuple!

plugins/modules/rds_cluster_info.py
chynasan: Since for rds_cluster_info, the only two API calls in rds.py are describe_db_clusters and list_tags_for_resource which both have their own @AWSretry decorators right now defined  I think that might double up on those retries?

Do you think we should re-add the retry_decorator here even though it's potentially redundant? Since other client calls in rds.py might not have those decorators added in the future for whatever reason, I can understand how this might be a good way to catch that. 

Happy to add it back if you prefer!

mjohns91:
I think if you rebase this on top of main, it should resolve your failing integration test! My recently merged [PR#3033](https://github.com/ansible-collections/amazon.aws/pull/3033/changes/9833bd485a69f35d585cb33a94299ed8f591279a) fixed the mariaDB issue.

oraNod (approved):
Thanks @chynasan

alinabuzachis (commented):
@chynasan Can you please use/update these new utils https://github.com/ansible-collections/amazon.aws/pull/3033?

beeankha (approved):

mjohns91:
I have rebased this work on top of the latest [rds refactor work](https://github.com/ansible-collections/amazon.aws/pull/3033) and combined the rds_cluster_info PR with this rds_cluster module work as part of that rebase. [PR#3040](https://github.com/ansible-collections/amazon.aws/pull/3040) is now technically included in this PR, so this will take care of the pair of modules and can be closed once this merges.

alinabuzachis (approved):

PR #3055, opened 2026-08-03
ci(integration.yml): add strip to pre-test-cmd scalar

mjohns91:
## Summary

- Fix YAML block scalar (`|`) to block-strip scalar (`|-`) for the `pre-test-cmd` input passed to `ansible-test-gh-action` in the integration workflow
- The `|` scalar includes a trailing newline, which the action's `format('set -x; {0}; set +x', ...)` call embeds into the generated shell script, producing a bare `; set +x` on its own line — a bash syntax error that fails all integration jobs
- Also restores the missing trailing newline at end of file

### Root cause

`ansible-test-gh-action` wraps the `pre-test-cmd` value with `format('set -x; {0}; set +x', inputs.pre-test-cmd)`. When the input contains a trailing newline (from YAML `|`), the generated script becomes:

```bash
set -x; python3 -m pip install -r tests/integration/requirements.txt -r tests/integration/constraints.txt
; set +x
```

The bare `; set +x` is a bash syntax error (`syntax error near unexpected token ';'`).

### Fix

Changing `|` to `|-` strips the trailing newline while keeping the block scalar style — the minimal change from the original.

### Affected runs

All integration jobs fail with this error — example: https://github.com/ansible-collections/amazon.aws/actions/runs/30841816931/job/91780607096?pr=3053

## Test plan

- [x] Verify integration CI passes on this PR (the workflow change itself is exercised)
- [ ] Backport to stable-10 and stable-11

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -164,5 +164,5 @@ jobs:
             community.crypto
             ansible.windows
             ${{ steps.set-path.outputs.collection_dep }}
-          pre-test-cmd: |
+          pre-test-cmd: >-
             python3 -m pip install -r tests/integration/requirements.txt -r tests/integration/constraints.txt
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -164,5 +164,5 @@ jobs:
             community.crypto
             ansible.windows
             ${{ steps.set-path.outputs.collection_dep }}
-          pre-test-cmd: >-
+          pre-test-cmd: |-
             python3 -m pip install -r tests/integration/requirements.txt -r tests/integration/constraints.txt


chynasan (approved):

.github/workflows/integration.yml
```diff
@@ -164,5 +164,5 @@ jobs:
             community.crypto
             ansible.windows
             ${{ steps.set-path.outputs.collection_dep }}
-          pre-test-cmd: |
+          pre-test-cmd: >-
             python3 -m pip install -r tests/integration/requirements.txt -r tests/integration/constraints.txt
```
oraNod: Since this is a single line I feel like the block scalar `>-` is unnecessary. You could use a plain scalar:

```
pre-test-cmd: python3 -m pip install -r tests/integration/requirements.txt -r tests/integration/constraints.txt
```

This is 121 characters long and is well inside the yamllint config of 160: https://github.com/mjohns91/amazon.aws/blob/efa6171cca7cde4dc56d6ea229e04a111a0302bb/.yamllint#L13

Maybe it's nitpicky but I think `>-` should be reserved for multiple lines folded into one, like `test-deps:` above  `pre-test-cmd:`.

oraNod:
I see there is a trailing new line at the end of file missing too.

image

This is different to the trailing new line from the `|` scalar, I believe. It's just a file-level convention and not part of the scalar value.

Looking at https://redocly.com/learn/yaml/blocks-and-flows#line-breaks-to-end-blocks

It's likely that we can simply add a strip like so `|-` to fix the issue and remove the trailing newline. I'd probably advocate for that to be honest as it is the minimal fix from the original. Just change `|` to `|-`.

tremble (approved):

mjohns91:
@oraNod Good catch. That is a better and simpler change. I have modified the approach to be that.

oraNod (approved):
Looks good. Thanks @mjohns91

oraNod:
@mjohns91 Sorry to be all nitpicky but I'd either drop the first commit or squash and rewrite so the commit message accurately reflects the change. Same with the PR title. I feel like that's a healthy habit that your future self might be grateful for. :wink:

tremble:
@oraNod Generally, when trigger a merge with mergeit it'll use the 1st comment for the description, not the commits, everything gets squashed into a single commit rather than merged as multiple commits.

beeankha (approved):

PR #3062, opened 2026-08-10
Refactor rds_engine_versions_info to use shared module_utils/rds utilities

mandar242:
##### SUMMARY

Refactor the `rds_engine_versions_info` module to use shared utility functions from `module_utils/_rds/`, consistent with the recent refactoring of `rds_cluster` (#3041) and the `module_utils/rds` decomposition (#3033).

Changes:
- Add shared `describe_db_engine_versions()` to `module_utils/_rds/api.py` with `@RDSErrorHandler.list_error_handler` and `@AWSRetry.jittered_backoff()`
- Re-export via `module_utils/rds.py` public facade
- Replace inline `@AWSRetry`-decorated function and bare `botocore` exception handling with shared API function and `AnsibleRDSError`
- Remove unnecessary `retry_decorator` on client creation and try/except wrapper
- Extract business logic into typed `engine_versions_info()` function with docstring
- Add unit tests for both the module (7 tests) and the shared API function (3 tests)

Relates to: https://redhat.atlassian.net/browse/ACA-2483

##### ISSUE TYPE

- Refactoring Pull Request
- Test Pull Request

##### COMPONENT NAME

rds_engine_versions_info

##### ADDITIONAL INFORMATION

This is part of the broader RDS refactoring effort to consolidate shared logic into `module_utils/_rds/` and ensure all RDS modules follow consistent patterns for API calls, error handling, and response transformation.

The refactoring follows the same patterns established in:
- PR #3041 (`rds_cluster` refactor)
- PR #3033 (`module_utils/rds` decomposition into `_rds/` subpackage)

No functional changes, the module behavior and return values are unchanged.

Assisted-by: Claude Opus 4.6

PR #3064, opened 2026-08-14
refactor(rds): rds_option_group uses shared module_utils/rds utilities

chynasan:
##### SUMMARY

Refactor `rds_option_group_info` to use shared utilities from `module_utils/rds`, following the pattern established by the `rds_cluster_info` refactor in #3041.

  - Add `describe_option_groups()` to `module_utils/_rds/api.py`
    with `@RDSErrorHandler` and `@AWSRetry` decorators
  - Add `OptionGroupNotFoundFault` to `RDSErrorHandler._is_missing()`
  - Re-export `describe_option_groups` via the `module_utils/rds` facade
  - Replace local `_describe_option_groups()` with the shared function
  - Standardize error handling with `AnsibleRDSError` caught in `main()`
  - Simplify client creation (retry lives in the shared describe function)
  - Extract business logic into typed `option_group_info()` function
    with explicit parameters
  - Fix typo in marker validation: "minutes" removed from error message
  - Add unit tests for the module and the shared API function

Fixes ACA-2485

##### ISSUE TYPE

  - Refactoring Pull Request

##### COMPONENT NAME

  rds_option_group_info
  module_utils/rds

##### ADDITIONAL INFORMATION

This PR will not pass integration tests until #3065 is  merged, since it uses the shared `describe_option_groups()` utility from that PR.

  Reference PRs:
  - #3033 — decomposed `module_utils/rds.py` into `_rds/` subpackage
  - #3041 — `rds_cluster` refactor (pattern followed)
  - #3062 — `rds_engine_versions_info` refactor (review feedback incorporated:
    alphabetical ordering in `api.py`, `msg=` on `fail_json_aws`,
    changelog fragment included)

  Assisted-by: Claude Code / Opus 4.6 (Anthropic)


changelogs/fragments/3064_rds_option_group-refactor.yml
```diff
@@ -0,0 +1,4 @@
+minor_changes:
+  - rds_option_group - Refactored module to use shared ``describe_option_groups` and ``get_tags`` from ``module_utils/rds``, and standardized error handling (https://github.com/ansible-collections/amazon.aws/pull/3064).
```
oraNod:
```suggestion
  - rds_option_group - Refactored module to use shared ``describe_option_groups`` and ``get_tags`` from ``module_utils/rds``, and standardized error handling (https://github.com/ansible-collections/amazon.aws/pull/3064).
```

plugins/modules/rds_option_group.py
```diff
@@ -507,26 +501,57 @@ def compare_option_group(client, module):
     return to_be_added, to_be_removed
 
 
-def setup_option_group(client, module):
+def update_tags(client, module: AnsibleAWSModule, option_group: Dict[str, Any]) -> bool:
```
oraNod, line 504: I'm not 100% but it looks like some of this logic is duplicating what `_rds/tags` is doing: https://github.com/chynasan/amazon.aws/blob/092de5f28e8fb60687fd40c4b0b5818b0b9a0e8e/plugins/module_utils/_rds/tags.py#L38

Maybe it could be simplied to something like:

```
def update_tags(client, module, option_group):
      existing_tags = get_tags(client, module, option_group["option_group_arn"])
      return ensure_tags(
          client,
          module,
          resource_arn=option_group["option_group_arn"],
          existing_tags=existing_tags,
          tags=module.params.get("tags"),
          purge_tags=module.params["purge_tags"],
      )
```

chynasan: The ensure_tags calls call_method() internally, which uses get_rds_method_attribute() to look up retry codes. That function raises NotImplementedError when wait=True and the resource type isn't registered in _RESOURCE_CONFIGS. Since rds_option_group has wait=True in its argument spec but option groups aren't in the resource configs, ensure_tags would fail at runtime. Modules that use ensure_tags either don't have a wait param (e.g., rds_subnet_group) or have their resource type registered (e.g., rds_cluster).

In order to add option groups to the resource configs, we'd need to modify  common.py, waiters.py and  rds.py at a minimum in the shared infrastructure to enable a one-liner ensure_tags call.

oraNod: OK, thanks for the explanation! I really wasn't sure if that suggestion was valid or not. TIL. Thanks @chynasan

oraNod (approved):
Thanks @chynasan 🎉

GomathiselviS (approved):

mandar242 (approved):

PR #3065, opened 2026-08-14
refactor(rds): rds_option_group_info uses shared module_utils/rds utilities

chynasan:
##### SUMMARY

Refactor `rds_option_group` to use shared utilities from `module_utils/rds`, following the pattern established by the `rds_cluster` refactor in #3041.

  - Replace local `_describe_option_groups()` with shared
    `describe_option_groups()` from `module_utils/rds`
  - Use shared `get_tags()` for tag retrieval instead of raw
    `list_tags_for_resource()` + manual conversion
  - Standardize error handling with `AnsibleRDSError` caught in `main()`
  - Preserve retry behavior on all direct client calls with
    `retry_decorator` and `aws_retry=True`
  - Clean up `update_tags()` using `get_tags` and `compare_aws_tags`,
    with proper check_mode early return before any boto3 call
  - Add type hints and docstrings to all functions
  - Fix return type annotation on `compare_option_group` to reflect
    `Optional` returns
  - Add consistent `msg=` to all `fail_json_aws` calls
  - Add unit tests covering `get_option_group`, `update_tags`,
    `create_option_group` check_mode, and `main()` success/failure paths

  Fixes ACA-2484

##### ISSUE TYPE

  - Bugfix Pull Request
  - Refactoring Pull Request

##### COMPONENT NAME

rds_option_group

##### ADDITIONAL INFORMATION

  **Bugfix:** `check_mode` was not respected during option group creation when tags were provided. The `if module.check_mode: return changed` guard was incorrectly nested inside the `else` branch of the tags
  conditional (only applied when tags were *not* provided), so `create_option_group` would proceed with the actual AWS API call in check_mode when tags were present.

 **Note on `ensure_tags`:** This module keeps tag handling inline rather than using the shared `ensure_tags()` utility. `ensure_tags` calls `call_method()`, which calls `get_rds_method_attribute()` - that function raises `NotImplementedError` when `wait=True` and  the resource type is not registered in `_RESOURCE_CONFIGS`. Since `rds_option_group` has `wait=True` but option groups are not in the resource configs, `ensure_tags` cannot be used without scope expansion. Modules without `wait` (e.g., `rds_subnet_group`) or with registered resource configs (e.g., `rds_cluster`) use `ensure_tags` successfully.

The shared describe_option_groups() utility is needed before CI will pass on the rds_option_group refactor PR #3064, so this PR will need to be merged first. That PR should then be able to pass integration tests.

Reference PRs:
  - #3041: `rds_cluster` refactor (pattern followed)
  - #3062: `rds_engine_versions_info` refactor (review feedback incorporated)

  Assisted-by: Claude Code / Opus 4.6 (Anthropic)


plugins/modules/rds_option_group_info.py
```diff
@@ -310,15 +321,23 @@ def main():
...
+                module,
+                option_group_name=module.params.get("option_group_name"),
+                engine_name=module.params.get("engine_name"),
+                major_engine_version=module.params.get("major_engine_version"),
+                marker=module.params.get("marker"),
+                max_records=module.params.get("max_records"),
+            ),
+        )
+    except AnsibleRDSError as e:
+        module.fail_json_aws(e, msg="Couldn't describe option groups.")
```
oraNod:
```suggestion
        module.fail_json_aws(e, msg="Could not describe option groups.")
```

oraNod: Kind of a nit but I'd avoid contractions like this for clarity.

plugins/modules/rds_option_group_info.py
```diff
@@ -235,58 +235,69 @@
...
+        params["MajorEngineVersion"] = major_engine_version
 
-    if module.params.get("max_records"):
-        params["MaxRecords"] = module.params.get("max_records")
-        if params["MaxRecords"] > 100:
-            module.fail_json(msg="The maximum number of records to include in the response is 100.")
+    if marker:
+        params["Marker"] = marker
+        if int(marker) < 20 or int(marker) > 100:
+            module.fail_json(msg="marker must be between 20 and 100")
```
oraNod: Should the message be a complete sentence for consistency with `MaxRecords` below?

plugins/modules/rds_option_group_info.py
```diff
@@ -235,58 +235,69 @@
...
+    if major_engine_version:
+        params["MajorEngineVersion"] = major_engine_version
 
-    if module.params.get("max_records"):
-        params["MaxRecords"] = module.params.get("max_records")
-        if params["MaxRecords"] > 100:
-            module.fail_json(msg="The maximum number of records to include in the response is 100.")
+    if marker:
+        params["Marker"] = marker
+        if int(marker) < 20 or int(marker) > 100:
```
oraNod: In the arg spec I see `marker=dict(type="str")` so if the user passes a string won't this result in an error?

For example:

```
marker = "mystring"
if int(marker) < 20 or int(marker) > 100:
  print("must be between 20 and 100")
```

This results in:

```
$ python marker.py 
  File "/home/dnaro/workspace/marker.py", line 2, in <module>
    if int(marker) < 20 or int(marker) > 100:
       ~~~^^^^^^^^
ValueError: invalid literal for int() with base 10: 'mystring'
```

While:

```
marker = "12"
if int(marker) < 20 or int(marker) > 100:
  print("must be between 20 and 100")
```

Results in:

```
$ python marker.py 
must be between 20 and 100
```

oraNod: OK, so I looked and see that marker is a string in AWS docs here: https://docs.aws.amazon.com/boto3/latest/reference/services/rds/client/describe_db_instances.html#

The `if int(marker) < 20 or int(marker) > 100:` validation should apply to `MaxRecords` not `marker`?

plugins/modules/rds_option_group_info.py
```diff
@@ -235,58 +235,69 @@
...
+            module.fail_json(msg="marker must be between 20 and 100")
 
-    params["EngineName"] = module.params.get("engine_name")
-    params["MajorEngineVersion"] = module.params.get("major_engine_version")
+    if max_records:
+        params["MaxRecords"] = max_records
+        if max_records > 100:
+            module.fail_json(msg="The maximum number of records to include in the response is 100.")
```
beeankha:
```suggestion
        if max_records < 20 or max_records > 100:
            module.fail_json(msg="The maximum number of records to include in the response must be between 20 and 100.")

params["MaxRecords"] = max_records            
```

plugins/modules/rds_option_group_info.py
beeankha, line 31:
```suggestion
            - The maximum number of records to include in the response.
            - Allowed values are between V(20) and V(100).
```

tests/unit/plugins/modules/test_rds_option_group_info.py
```diff
@@ -0,0 +1,134 @@
+# (c) 2026 Red Hat Inc.
```
beeankha, line 1: It might be good to have test coverage here for `max_records` value being too high and/or too low.

beeankha (approved):

oraNod (approved):
LGTM. Thanks @chynasan 👍

PR #3042, opened 2026-07-27
fix: normalize telemetry taxonomy device_type and infra_bucket values

stevefulme1:
## Summary

The `event_query.yml` audit filter emits vendor-specific strings as `device_type` values and uses inconsistent casing for `infra_bucket` / `infra_type`. This blocks automated cross-vendor reporting against the Ansible Normalized Resource Taxonomy v1.0.

**Before:**
- `device_type: "EC2"`, `"ApplicationLoadBalancer"`, `"Bucket"`, `"Route table"`, `"NAT Gateway"`, `"Instance"`, etc.
- `infra_bucket: "Compute"`, `"Networking"`, `"Storage"`, `"Database"`
- `infra_type: "PublicCloud"`

**After:**
- `device_type: "virtual_machine"`, `"load_balancer"`, `"object_storage"`, `"router"`, `"nat_gateway"`, `"database_relational"`, etc.
- `infra_bucket: "compute"`, `"networking"`, `"storage"`, `"database"`
- `infra_type: "public_cloud"`

### Normalization Map

All 15 module queries updated:

| Module | Old `device_type` | New `device_type` |
|---|---|---|
| `ec2_instance_info` | `EC2` | `virtual_machine` |
| `elb_application_lb_info` | `ApplicationLoadBalancer` | `load_balancer` |
| `elb_classic_lb_info` | `ClassicLoadBalancer` | `load_balancer` |
| `ec2_vpc_net_info` | `VPC` | `vpc` |
| `ec2_vpc_subnet_info` | `Subnet` | `vpc` |
| `ec2_vpc_nat_gateway_info` | `NAT Gateway` | `nat_gateway` |
| `ec2_vpc_igw_info` | `Internet Gateway` | `internet_gateway` |
| `ec2_vpc_vgw_info` | `Virtual Gateway` | `vpn_gateway` |
| `ec2_vpc_route_table_info` | `Route table` | `router` |
| `ec2_vpc_vpn_info` | `VPN Connection` | `vpn_gateway` |
| `ec2_vpc_peering_info` | `VPC Peering` | `vpc_peering` |
| `s3_bucket_info` | `Bucket` | `object_storage` |
| `s3_object_info` | `Object` | `object_storage` |
| `rds_instance_info` | `Instance` | `database_relational` |
| `rds_cluster_info` | `Cluster` | `database_relational` |

[...]

PR #3066, opened 2026-08-14
event_query: emit a non-null top-level name for every module

fistilis:
##### SUMMARY

Every module's `extensions/audit/event_query.yml` entry should emit a **non-null top-level `name`** for the indirect-node audit path, but currently only `ec2_instance_info` and `ec2_vpc_net_info` do. The other **13 modules** (all networking, storage, and database `*_info` queries) place `name` only under `canonical_facts` or omit it entirely.

The controller persists each jq result into `main_indirectmanagednodeaudit`, whose `name` column is **`NOT NULL`**, using a single `bulk_create`. A record with a null top-level `name` raises an integrity error that aborts the **entire** statement — so a job that gathers even one such resource (e.g. a subnet, S3 bucket, or RDS instance) silently discards **all** of the indirect nodes it counted, while the job still reports success.

This PR adds a top-level `name` to each of the 13 remaining modules, mirroring the existing `(.name // .tags.Name // "UNKNOWN")` fallback already used by `ec2_instance_info`/`ec2_vpc_net_info`, deriving the value from the resource's own tag/identifier so it is never null.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

event_query (`extensions/audit/event_query.yml`)

##### ADDITIONAL INFORMATION

**Modules fixed** (13): `elb_application_lb_info`, `elb_classic_lb_info`, `ec2_vpc_subnet_info`, `ec2_vpc_nat_gateway_info`, `ec2_vpc_igw_info`, `ec2_vpc_vgw_info`, `ec2_vpc_route_table_info`, `ec2_vpc_vpn_info`, `ec2_vpc_peering_info`, `s3_bucket_info`, `s3_object_info`, `rds_instance_info`, `rds_cluster_info`.

Example (subnet):
```diff
     .subnets[] | select(. != null) | {
+      name: (.tags.Name // .id // "UNKNOWN"),  # unique identifier for the resource
       canonical_facts: {
         id: .id,
```

**Tests:** the `node_query_*` integration targets are updated so their expected results / templates now assert the new top-level `name` (networking templates gain a `name:` line matching `vpcs.j2`'s existing precedent; the storage/database/elb targets assert it directly, including the existing no-tags fallback cases).

**Verification:** the jq for all 15 modules was validated offline against representative `*_info` fixtures (tagged and untagged resources) confirming a non-null top-level `name` in every case; the same check fails on `main` for the 13 offending modules. The live-AWS integration run is left to CI.

**Related:** this is the collection half of a two-part fix — the controller should additionally be resilient to a null `name` (skip/derive rather than abort the whole batch) so collections that ship no `event_query.yml` are also protected; that is being tracked separately on the controller side. Also touches the same file as the in-flight taxonomy work in #3042 (cc @stevefulme1) — no overlap in the lines changed.

### DIFF
--- a/extensions/audit/event_query.yml
+++ b/extensions/audit/event_query.yml
@@ -4,7 +4,7 @@
 amazon.aws.ec2_instance_info:
   query: >-
     .instances[] | select(. != null) | {
-      name: (.name // .tags.Name // "UNKNOWN"),  # unique identifier for the resource
+      name: (.name // .tags.Name // .instance_id // "UNKNOWN"),  # unique identifier for the resource
       canonical_facts: {
       id: .instance_id,
       status: .state.name,
@@ -23,6 +23,7 @@ amazon.aws.elb_application_lb_info:
   query: >-
     .load_balancers[] | select(. != null) |
     {
+      name: (.load_balancer_name // .tags.Name // "UNKNOWN"),  # unique identifier for the resource
       canonical_facts: {
         name: .load_balancer_name,
         status: .state.code,
@@ -39,6 +40,7 @@ amazon.aws.elb_classic_lb_info:
   query: >-
     .elbs[] | select(. != null) |
     {
+      name: (.load_balancer_name // .tags.Name // "UNKNOWN"),  # unique identifier for the resource
       canonical_facts: {
         name: .load_balancer_name,
         tags: (.tags // {}),
@@ -53,7 +55,7 @@ amazon.aws.elb_classic_lb_info:
 amazon.aws.ec2_vpc_net_info:
   query: >-
     .vpcs[] | select(. != null) | {
-        name: (.name // .tags.Name // "UNKNOWN"),
+        name: (.name // .tags.Name // .id // "UNKNOWN"),
         canonical_facts: {
           id: .id,
           status: .state,
@@ -69,6 +71,7 @@ amazon.aws.ec2_vpc_net_info:
 amazon.aws.ec2_vpc_subnet_info:
   query: >-
     .subnets[] | select(. != null) | {
+      name: (.tags.Name // .id // "UNKNOWN"),  # unique identifier for the resource
       canonical_facts: {
         id: .id,
         status: .state,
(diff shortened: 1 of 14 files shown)


alinabuzachis (commented):
@fistilis Thank you for this work. Could you please update the documentation to reflect the new changes https://github.com/ansible-collections/amazon.aws/blob/main/extensions/audit/README.md? Thanks.

extensions/audit/event_query.yml
```diff
@@ -69,6 +71,7 @@ amazon.aws.ec2_vpc_net_info:
 amazon.aws.ec2_vpc_subnet_info:
   query: >-
     .subnets[] | select(. != null) | {
+      name: (.tags.Name // .id // "UNKNOWN"),  # unique identifier for the resource
```
alinabuzachis, line 74: Should we consider .id too if .tags.Name is not set?

fistilis: Good call — done. I made the identifier fallback consistent across the whole file: `ec2_instance_info` now uses `.name // .tags.Name // .instance_id // "UNKNOWN"` and `ec2_vpc_net_info` uses `.name // .tags.Name // .id // "UNKNOWN"` (every tag-based networking resource already fell back to its `.<resource>_id`). So an unnamed, untagged resource is now labelled by its ID instead of collapsing to `UNKNOWN`.

fistilis:
@alinabuzachis Thanks for the review — both points addressed in the latest push:

1. **README** ([`extensions/audit/README.md`](https://github.com/ansible-collections/amazon.aws/blob/main/extensions/audit/README.md)): added a **Top-Level `name`** section documenting the new required, non-null top-level `name`, why it's mandatory (controller column is `NOT NULL`), and the per-resource fallback chain.
2. **`.id` fallback**: made the identifier fallback consistent across the file (`ec2_instance_info` → `.instance_id`, `ec2_vpc_net_info` → `.id`) so unnamed/untagged resources are labelled by ID rather than `UNKNOWN`. Details in the inline reply.

Rebased onto latest `main`; commit is SSH-signed (GitHub verified). Ready for another look.

fistilis:
Closing this in favour of #3092, which is a strict superset.

#3066 fixed the missing top-level `name`. A fuller audit of this query file against everything `awx/main/tasks/host_indirect.py` requires found three further defect classes in the same fifteen queries — two jq errors that discard every module's records for the whole event, volatile `status`/`tags` fields in the sole dedup key, and unnormalized taxonomy on all fifteen. Splitting those across two PRs against one file only creates conflicts, so #3092 carries the lot, plus a static contract test.

@alinabuzachis — thank you for the review here; both of your points are carried over to #3092.

PR #3072, opened 2026-08-21
refactor(rds_subnet_group): use shared module_utils/rds utilities

mandar242:
##### SUMMARY
Replace the module-local backoff/describe and duplicated try/except blocks with the shared describe_db_subnet_groups and call_method helpers, and standardize error handling on AnsibleRDSError. Add describe_db_subnet_groups to _rds/api.py, DBSubnetGroupNotFoundFault to RDSErrorHandler, unit tests, and a changelog fragment.

##### ISSUE TYPE

- Refactoring Pull Request

##### COMPONENT NAME

rds_subnet_group

##### ADDITIONAL INFORMATION

Assisted-by: Claude Opus 4.8

PR #3074, opened 2026-08-26
fix: address failing sanity test checks from trailing newline

oraNod:
##### SUMMARY

This change addresses the following `ansible-test sanity` check failure:

```
ERROR: Found 1 pylint issue(s) which need to be resolved:
ERROR: tests/unit/plugins/module_utils/rds/test_rds_api.py:717:0: trailing-newlines: Trailing newlines
```

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/tests/unit/plugins/module_utils/rds/test_rds_api.py
+++ b/tests/unit/plugins/module_utils/rds/test_rds_api.py
@@ -714,4 +714,3 @@ class TestDescribeDbEngineVersions:
             Engine="postgres", DefaultOnly=True, DBParameterGroupFamily="postgres16",
         )
         assert len(result) == 1
-


plugins/modules/rds_option_group.py
```diff
@@ -351,8 +351,11 @@
 from typing import Optional
 from typing import Tuple
 
-from botocore.exceptions import BotoCoreError
-from botocore.exceptions import ClientError
+try:
+    from botocore.exceptions import BotoCoreError
+    from botocore.exceptions import ClientError
+except ImportError:
+    pass
```
alinabuzachis: I do not think this should be added back. Instead, errors should have been handled differently. Look for example https://github.com/ansible-collections/amazon.aws/pull/3072.

alinabuzachis (changes requested):

plugins/modules/rds_option_group.py
tremble: +1 the general direction has been to move the client call into a small function (with just the call), and use one of the RDSErrorHandler decorators to catch the errors.  Then you can completely drop botocore imports from here.

oraNod: Thanks @alinabuzachis and @tremble I guess it was kind of stoopid to add the try/except block back in after all. I see how it should work now. Apologies for the noise. I think I just got zoned in on the sanity test failure rather than looking at the refactoring PRs. 

Anyway I've dropped that commit from this PR and will just leave it open to sort the trailing newline. Cheers.

GomathiselviS (approved):

alinabuzachis (approved):

abikouo (approved):

chynasan (approved):

oraNod:
Fixed on `main` with https://github.com/ansible-collections/amazon.aws/pull/3072 so I'll close this PR

PR #3078, opened 2026-09-02
Refactor rds_option_group drops direct botocore imports for RDSErrorHandler

chynasan:
##### SUMMARY
 `rds_option_group.py` currently imports `botocore.exceptions.BotoCoreError`/`ClientError` directly and wraps every client call (create/modify/delete option group, add/remove tags) in its own local `try`/`except`. A prior attempt to fix a sanity failure related to these imports (#3074) patched the symptom by wrapping the imports in `try/except ImportError`, but reviewer feedback (@tremble) pointed out the correct direction: move each client call into a small function decorated with an `RDSErrorHandler` decorator so the module can drop the direct `botocore` imports entirely, matching the pattern already used in `rds_subnet_group` (#3072).
 
This PR:
- Removes the direct `botocore.exceptions` imports and all inline `try/except (ClientError, BotoCoreError)` blocks.
- Adds five small private helper functions (`_modify_option_group`, `_remove_option_group_options`, `_create_option_group`, `_delete_option_group`, `_add_tags_to_resource`, `_remove_tags_from_resource`) decorated with `RDSErrorHandler.common_error_handler`/`RDSErrorHandler.deletion_error_handler`, which raise `AnsibleRDSError`, already caught centrally in `main()`.
- Deliberately does **not** reuse the shared `call_method`/`ensure_tags` helpers from `module_utils/rds`: those route through `get_rds_method_attribute`, which raises `NotImplementedError` for methods not registered in `_RESOURCE_CONFIGS` when `module.params["wait"]` is truthy. `rds_option_group` defaults `wait=True` (though `wait` has no effect in this module today — no boto3 waiter exists for option groups), and none of its client methods are registered for that resource type, so reusing `call_method` as-is would have broken every write operation.
- Fixes `remove_option_group` to derive `changed` from the delete call's actual result instead of pre-setting `changed = True`, so a concurrent "already deleted" race correctly reports `changed: false`.
- Gives the remove-options call site its own error description ("remove options from option group") instead of sharing the generic "update option group" message.
- Adds a regression test (`test_remove_option_group_already_absent`) covering the not-found/no-op delete path. 
 
Verified locally: `ansible-test units --python 3.14` (676 module tests / 1522 module_utils tests passing), `ansible-test sanity --python 3.14 plugins/modules/rds_option_group.py` (clean), and `ansible-test  integration --python 3.14 rds_option_group` against real AWS (52 tasks ok, 0 failed).
 
##### ISSUE TYPE
- Refactoring Pull Request
  
##### COMPONENT NAME
  rds_option_group
  
##### ADDITIONAL INFORMATION
Follow-up to #3074 and companion to the `rds_subnet_group` refactor in #3072.
 
Assisted-by: Claude Code / Sonnet 5 (Anthropic)

### DIFF
--- a/changelogs/fragments/rds_option_group-drop-botocore-imports.yml
+++ b/changelogs/fragments/rds_option_group-drop-botocore-imports.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - rds_option_group - Removed direct ``botocore.exceptions`` imports and inline ``try``/``except`` blocks; option group create, modify, delete, and tagging calls are now made through small helper functions decorated with ``RDSErrorHandler``, standardizing error handling on ``AnsibleRDSError`` (https://github.com/ansible-collections/amazon.aws/pull/3074).
+  - rds_option_group - Removed direct ``botocore.exceptions`` imports and inline ``try``/``except`` blocks; option group create, modify, delete, and tagging calls are now made through small helper functions decorated with ``RDSErrorHandler``, standardizing error handling on ``AnsibleRDSError`` (https://github.com/ansible-collections/amazon.aws/pull/3078).
--- a/changelogs/fragments/rds_option_group-drop-botocore-imports.yml
+++ b/changelogs/fragments/rds_option_group-drop-botocore-imports.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - rds_option_group - Removed direct ``botocore.exceptions`` imports and inline ``try``/``except`` blocks; option group create, modify, delete, and tagging calls are now made through small helper functions decorated with ``RDSErrorHandler``, standardizing error handling on ``AnsibleRDSError`` (https://github.com/ansible-collections/amazon.aws/pull/3074).
--- a/changelogs/fragments/rds_option_group-fix-remove-race-changed.yml
+++ b/changelogs/fragments/rds_option_group-fix-remove-race-changed.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - rds_option_group - ``state=absent`` no longer fails when the option group is deleted by another process between the existence check and the delete call; ``changed`` is now derived from the delete call's actual result instead of being pre-set to ``True`` (https://github.com/ansible-collections/amazon.aws/pull/3078).
--- a/plugins/modules/rds_option_group.py
+++ b/plugins/modules/rds_option_group.py
@@ -351,19 +351,47 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 
-from botocore.exceptions import BotoCoreError
-from botocore.exceptions import ClientError
-
 from ansible.module_utils.common.dict_transformations import camel_dict_to_snake_dict
 from ansible.module_utils.common.dict_transformations import snake_dict_to_camel_dict
 
 from ansible_collections.amazon.aws.plugins.module_utils.modules import AnsibleAWSModule
 from ansible_collections.amazon.aws.plugins.module_utils.rds import AnsibleRDSError
+from ansible_collections.amazon.aws.plugins.module_utils.rds import RDSErrorHandler
 from ansible_collections.amazon.aws.plugins.module_utils.rds import describe_option_groups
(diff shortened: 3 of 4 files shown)


changelogs/fragments/rds_option_group-drop-botocore-imports.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - rds_option_group - Removed direct ``botocore.exceptions`` imports and inline ``try``/``except`` blocks; option group create, modify, delete, and tagging calls are now made through small helper functions decorated with ``RDSErrorHandler``, standardizing error handling on ``AnsibleRDSError`` (https://github.com/ansible-collections/amazon.aws/pull/3078).
```
GomathiselviS, line 2: The PR fixes a bug (a delete racing with a concurrent deletion previously caused a hard module failure via fail_json_aws; now it correctly reports changed: false). Users scanning the changelog for behavior changes won't see this fix documented. Can we add a "bugfixes" changelog fragment?

GomathiselviS (approved):
Please add the missing bugfix changelog fragment. The PR looks good to me otherwise!

changelogs/fragments/rds_option_group-drop-botocore-imports.yml
tremble: Was that bug actually released?

chynasan: I asked Claude to dig around for the origin of this and it seems like the bug that changed pre-set to True, hard fail via fail_json_aws on a delete-race with concurrent deletion has been present since at least the 6.0.0 release, so I'm gonna push that changelog fragment now. Good to check though! :)

GomathiselviS (approved):

changelogs/fragments/rds_option_group-fix-remove-race-changed.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
abikouo: It is fine like that, but both changes could have been merged in the same file

plugins/modules/rds_option_group.py
```diff
@@ -351,21 +351,49 @@
...
 
 
+@RDSErrorHandler.common_error_handler("update option group")
+def _modify_option_group(client, **params: Any) -> Dict[str, Any]:
+    return client.modify_option_group(aws_retry=True, **params)
+
+
+@RDSErrorHandler.common_error_handler("remove options from option group")
+def _remove_option_group_options(client, **params: Any) -> Dict[str, Any]:
+    return client.modify_option_group(aws_retry=True, **params)
```
abikouo: Both functions are calling the same botocore API; is this expected? Can we merge this?

plugins/modules/rds_option_group.py
```diff
@@ -412,10 +437,7 @@ def remove_option_group_options(client, module: AnsibleAWSModule, options_to_rem
     if module.check_mode:
         return changed
 
-    try:
-        client.modify_option_group(aws_retry=True, **params)
-    except (ClientError, BotoCoreError) as e:
-        module.fail_json_aws(e, msg="Unable to remove options from Option Group.")
+    _remove_option_group_options(client, **params)
```
abikouo:
```suggestion
    _modify_option_group(client, **params)
```

abikouo (changes requested):
Looks good overall, just some remarks

plugins/modules/rds_option_group.py
chynasan: This is really good feedback! I poked at it a little for my own benefit and you're right, RDSErrorHandler should output a detailed enough message so I'll merge them.  Thanks!

changelogs/fragments/rds_option_group-fix-remove-race-changed.yml
chynasan: I merged them anyway. :P

abikouo (approved):

mandar242 (approved):

PR #3082, opened 2026-09-03
test(module_utils): populate exception field on ansible-core 2.19

### LINKED ISSUE #2929: fail_json_aws() exception field not populated in ansible-core >= 2.19
### Summary

The `exception` field is not populated in the JSON output from `fail_json_aws()` when running with ansible-core >= 2.19, causing test failures in `tests/unit/plugins/module_utils/modules/ansible_aws_module/test_fail_json_aws.py`.

### Issue Type

Bug Report

### Component Name

`AnsibleAWSModule.fail_json_aws()`

### Ansible Version

ansible-core >= 2.19 (stable-2.19, stable-2.20, milestone, devel)

### Collection Version

Current main branch (12.0.0-dev0)

### Configuration

N/A

### OS / Environment

All platforms - discovered during CI migration to ansible-test-gh-action

### Steps to Reproduce

1. Run unit tests with ansible-core 2.19 or later:
   ```bash
   tox -e ansible2.19-py312-with_constraints -- tests/unit/plugins/module_utils/modules/ansible_aws_module/test_fail_json_aws.py -v
   ```

2. All 10 tests in `TestFailJsonAwsTestSuite` fail with:
   ```python
   assert return_val.get("exception") is not None
   AssertionError: assert None is not None
   ```

### Expected Results

The `exception` field should contain the formatted traceback when `fail_json_aws()` is called, as it does in ansible-core < 2.19.

### Actual Results

The `exception` field is `None` (not present in the JSON output).

Example failing test output:
```python
return_val = {
    'msg': 'An unspecified error occurred',
    'boto3_version': '1.2.4',
    'botocore_version': '1.2.3',
    'failed': True,
    'deprecations': [...],
    # 'exception' field is missing
}
```

### Additional Context

This appears to be related to changes in ansible-core 2.19's exception handling in `fail_json()`. The deprecation message "Passing `warnings` to `exit_json` or `fail_json` is deprecated" also appears in the output.

**Tests affected (10 total):**
- `test_fail_client_minimal`
- `test_fail_client_msg`
- `test_fail_client_positional_msg`
- `test_fail_client_key`
- `test_fail_client_msg_and_key`
- `test_fail_botocore_minimal`
- `test_fail_botocore_msg`
- `test_fail_botocore_positional_msg`
- `test_fail_botocore_key`
- `test_fail_botocore_msg_and_key`

**Relevant code:**
- `plugins/module_utils/modules.py:207-239` (`fail_json_aws()` method)
- Sets `exception=last_traceback` in the failure dict (line 232)

**References:**
- Ansible-core 2.19 porting guide: https://docs.ansible.com/projects/ansible/latest/porting_guides/porting_guide_core_2.19.html
- Discovered during: #2927 (CI migration to ansible-test-gh-action)

### Possible Solutions

1. **Conditional field name**: Use `error_traceback` instead of `exception` for ansible-core >= 2.19
2. **Dual fields**: Populate both `exception` and `error_traceback` for compatibility
3. **Update tests only**: Mark tests as xfail for ansible-core >= 2.19 until upstream resolution
4. **Investigate upstream**: Determine if this is an ansible-core regression and report upstream

**Temporary workaround:**
Tests have been marked with `pytest.xfail()` for ansible-core >= 2.19 in the CI migration PR to unblock CI while we investigate the proper fix.

mandar242:
##### SUMMARY
Fixes #2929

The TestFailJsonAwsTestSuite unit tests were failing (and had been temporarily marked xfail) on ansible-core >= 2.19 because the exception field was no longer populated in the fail_json_aws() output.

This is an intentional, by-design change in ansible-core 2.19: the target/module side now only reports a traceback (the exception field) when it has been explicitly enabled for the module, driven by the _ansible_tracebacks_for module argument,  not by the controller-side ANSIBLE_DISPLAY_TRACEBACK setting the tests were relying on (which is a no-op module-side). 

AnsibleAWSModule.fail_json_aws() itself is already correct across all supported cores, so the fix is test-only.

- Enable module-side traceback reporting in the tests via _ansible_tracebacks_for: ["error"], version-gated so it is only passed on ansible-core >= 2.19 (older cores reject it as an unsupported parameter and populate exception unconditionally anyway).
- Add an autouse fixture that resets ansible-core's process-global traceback cache (_traceback._module_tracebacks_enabled_events) between tests, so enabling tracebacks in one test cannot leak into the negative-case tests. Guarded so pre-2.19 cores (no such module) skip it.
- Remove the misleading ANSIBLE_DISPLAY_TRACEBACK fixture and the 10 pytest.xfail(...) workarounds, restoring the real assert return_val.get("exception") is not None assertions.

No changelog fragment is included, as this is a test-only change (module behavior is unchanged).

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
module_utils/modules


tremble (approved):

tremble:
test failures are from RDS quota problems

mandar242:
recheck

abikouo (approved):

tests/unit/plugins/module_utils/modules/ansible_aws_module/test_fail_json_aws.py
```diff
@@ -33,11 +33,14 @@ def _ansible_version_at_least(version_str):
 if not HAS_BOTO3:
     pytestmark = pytest.mark.skip("test_fail_json_aws.py requires the python modules 'boto3' and 'botocore'")
 
-
-@pytest.fixture(autouse=True)
-def set_ansible_traceback_env(monkeypatch):
-    monkeypatch.setenv("ANSIBLE_DISPLAY_TRACEBACK", "always")
-    yield
```
oraNod, line 40: Looks like the environment variable is also present in https://github.com/ansible-collections/amazon.aws/blob/e37c6e76b8cbf061f4ab29dcc63483c926a1d1c4/tests/unit/plugins/module_utils/modules/ansible_aws_module/test_require_at_least.py#L48

tests/unit/plugins/module_utils/conftest.py
oraNod:
```suggestion
            request.param = {"ANSIBLE_MODULE_ARGS": request.param.copy()}
```

oraNod: Found a couple of issues with a Claude review:

1. stdin fixture mutates shared constant (conftest.py:60) - The fixture adds keys to
     request.param["ANSIBLE_MODULE_ARGS"] which is a reference to the module-level
     TRACEBACK_ARGS dict, permanently mutating it after the first test runs.
2. TRACEBACK_ARGS is mutable (test_fail_json_aws.py:43) - Shared dict defined once at module
     import is mutated by the stdin fixture, affecting all subsequent tests and preventing safe
     concurrent execution.

This suggested change ensures that:

  - Each test gets its own copy of the args dict
  - TRACEBACK_ARGS constant stays immutable
  - No cross-test pollution
  - Safe concurrent test execution

tests/unit/plugins/module_utils/modules/ansible_aws_module/test_fail_json_aws.py
mandar242: Good catch - removed that fixture (and the unused `os` import) too: it set `ANSIBLE_DISPLAY_TRACEBACK` directly on `os.environ` (leaking process-wide) and was a no-op either way, since module-side tracebacks come from `_ansible_tracebacks_for`, not the controller-side env var.

tests/unit/plugins/module_utils/conftest.py
mandar242: Applied - `request.param.copy()` is in place.

mandar242: Agreed on both - the `.copy()` fixes it, so `TRACEBACK_ARGS` is no longer mutated with `_ansible_remote_tmp`/`_ansible_keep_remote_files` by the first test that uses it.

oraNod (approved):
Thanks @mandar242

PR #3083, opened 2026-09-08
Add CODEOWNERS

mandar242:
##### SUMMARY

Add a CODEOWNERS file to ensure PRs are automatically assigned to the @ansible-collections/cloud team for review.

The CODEOWNERS file uses a wildcard (`*`) pattern to assign the cloud team as the default owner for all files in the repository. Validated syntax using codeowners-validator.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
.github/CODEOWNERS

PR #3084, opened 2026-09-09
rds_cluster - fix final_snapshot_identifier dropped on cluster deletion

### LINKED ISSUE #3060: rds_cluster: final_snapshot_identifier parameter silently dropped on delete
## Summary

The `rds_cluster` module's `final_snapshot_identifier` parameter does not work when deleting a cluster with `skip_final_snapshot: false`. The parameter name is converted incorrectly by `arg_spec_to_rds_params`, causing it to be silently dropped or rejected by boto3.

## Root Cause

`arg_spec_to_rds_params` converts `final_snapshot_identifier` → `FinalSnapshotIdentifier`, but boto3's `delete_db_cluster` expects `FinalDBSnapshotIdentifier` (with `DB`).

The `aws_replace_keys` fixup in `arg_spec_to_rds_params` replaces `Db` → `DB`, but `FinalSnapshotIdentifier` does not contain `Db`, so no replacement occurs.

## Comparison with rds_instance

`rds_instance.py` handles this correctly by using `final_db_snapshot_identifier` as the canonical parameter name (with `final_snapshot_identifier` as an alias):

```python
# plugins/modules/rds_instance.py:1672
final_db_snapshot_identifier=dict(aliases=["final_snapshot_identifier"]),
```

`rds_cluster.py` uses only `final_snapshot_identifier`:

```python
# plugins/modules/rds_cluster.py
final_snapshot_identifier=dict(),
```

## Impact

Attempting to delete a cluster with `skip_final_snapshot: false` and `final_snapshot_identifier` set will fail with an API error rather than creating the final snapshot. No data loss occurs — the delete operation fails — but the feature is broken.

## How to Reproduce

```yaml
- amazon.aws.rds_cluster:
    db_cluster_identifier: my-cluster
    state: absent
    skip_final_snapshot: false
    final_snapshot_identifier: my-final-snapshot
```

This will fail instead of creating the final snapshot and deleting the cluster.

## Suggested Fix

Rename the parameter to `final_db_snapshot_identifier` with `final_snapshot_identifier` as a deprecated alias, matching `rds_instance`:

```python
final_db_snapshot_identifier=dict(aliases=["final_snapshot_identifier"]),
```

Since `final_snapshot_identifier` has never worked for its intended purpose, the alias could potentially be removed without a deprecation cycle, but a deprecation is safer for users who have the parameter in their playbooks (even if the delete silently fails).

## Notes

- No integration test covers `skip_final_snapshot: false` — all existing tests use `skip_final_snapshot: true`
- Discovered during review of PR #3041

KR-Ravindra:
## Problem

Deleting a DB cluster with `skip_final_snapshot: false` and `final_snapshot_identifier` set fails: the snapshot identifier never reaches the API. The module passes the option validation, but AWS rejects the `DeleteDBCluster` call because `FinalDBSnapshotIdentifier` is missing while `SkipFinalSnapshot` is false.

## Root cause

- `plugins/modules/rds_cluster.py:1122` declares the option as `final_snapshot_identifier=dict()`.
- `plugins/module_utils/_rds/transformations.py:18` (`arg_spec_to_rds_params`) camel-cases it to `FinalSnapshotIdentifier`; the `("Db", "DB")` replacement never fires because the option name has no `db` segment.
- `plugins/module_utils/_rds/transformations.py:54` (`format_rds_client_method_parameters`) keeps only the keys that `delete_db_cluster` accepts, so the unknown `FinalSnapshotIdentifier` is silently dropped.
- `rds_instance` already avoids this: `plugins/modules/rds_instance.py:1672` declares `final_db_snapshot_identifier=dict(aliases=["final_snapshot_identifier"])`.

## Fix

Rename the option to `final_db_snapshot_identifier`, which camel-cases to `FinalDBSnapshotIdentifier` (the name `delete_db_cluster` expects), and keep `final_snapshot_identifier` as an alias, mirroring `rds_instance`. The DOCUMENTATION block, the `skip_final_snapshot` validation message and an example are updated, and a changelog fragment is added. Existing playbooks keep working through the alias.

## How tested

New unit test `test_main_delete_cluster_with_final_snapshot` in `tests/unit/plugins/modules/test_rds_cluster.py` drives `main()` with `state: absent`, validates the module arguments against the real argument spec with `ArgumentSpecValidator`, uses an offline botocore `rds` client so the parameter filter is exercised, and asserts on the parameters handed to `delete_db_cluster`. It is parametrized over the new option name and the alias.

Before the fix (`ansible-test units --python 3.12 tests/unit/plugins/modules/test_rds_cluster.py`):

```
>       assert delete_params["FinalDBSnapshotIdentifier"] == "my-cluster-final"
E       KeyError: 'FinalDBSnapshotIdentifier'
FAILED tests/unit/plugins/modules/test_rds_cluster.py::test_main_delete_cluster_with_final_snapshot[final_snapshot_identifier]
FAILED tests/unit/plugins/modules/test_rds_cluster.py::test_main_delete_cluster_with_final_snapshot[final_db_snapshot_identifier]
========================= 2 failed, 31 passed in 2.19s =========================
```

(The `final_db_snapshot_identifier` case fails before the fix with "Unsupported parameters", since the option did not exist yet.)

After the fix:

```
PASSED tests/unit/plugins/modules/test_rds_cluster.py::test_main_delete_cluster_with_final_snapshot[final_db_snapshot_identifier]
PASSED tests/unit/plugins/modules/test_rds_cluster.py::test_main_delete_cluster_with_final_snapshot[final_snapshot_identifier]
============================== 33 passed in 2.02s ==============================
```

Also run: `tests/unit/plugins/module_utils/rds/test_rds_transformations.py` (24 passed); `ansible-test sanity --test validate-modules --test pep8 --test pylint --test yamllint` on the changed files (clean); `antsibull-changelog lint` (clean); black, isort and flynt on the changed files (no changes from this PR; `rds_cluster.py` has a pre-existing black difference in unrelated function signatures that this PR leaves alone).

No integration test change: the `rds_cluster_*` integration targets all use `skip_final_snapshot: true`.

## Links

Fixes https://github.com/ansible-collections/amazon.aws/issues/3060

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
rds_cluster

##### ADDITIONAL INFORMATION

Assisted-by: Claude Fable 5.1 (claude-fable-5-1)

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.


KR-Ravindra:
Round 1 self-review.

1. The rename mirrors `rds_instance` (`final_db_snapshot_identifier=dict(aliases=["final_snapshot_identifier"])`, `plugins/modules/rds_instance.py:1672`) and camel-cases to `FinalDBSnapshotIdentifier`, the key `delete_db_cluster` accepts, so `format_rds_client_method_parameters` no longer drops it. Nothing else in the repo references the old `rds_cluster` option name.
2. `ansible-test units --python 3.12 tests/unit/plugins/modules/test_rds_cluster.py`: 33 passed at 0afb014; the new parametrized test fails on `main` with `KeyError: 'FinalDBSnapshotIdentifier'`. `test_rds_transformations.py`: 24 passed.
3. Changelog fragment present; DOCUMENTATION, argspec alias and the `skip_final_snapshot` message agree. The `changelog`, `all_green` and `update-variables` runs finish with no jobs on every open PR here at the moment (e.g. #3083), so those failures are not caused by this change; the Zuul check is green. Marking ready.

Issue #3063, opened 2026-08-12
tags element in ec2_tags is simultaneously required and deprecated


Synchro:
### Summary

The `ec2_tags` module _requires_ a property called `tags`, containing a list of tag names and values. However, it appears that ansible itself has now deprecated any module having a `tags` property, so I get a warning when it runs:

```
[WARNING]: Found variable using reserved name 'tags'.
Origin: <unknown>

tags
```

There is no mention of this deprecation in the ec2_tags module docs, nor is there an alias for it like there is in ec2_instance (`resource_tags`). I don't know why the origin appears as `<unknown>`, but this is the only place in my entire ansible setup that uses the term `tags`.

I also get the warning in #3028, but this is a different issue.

### Issue Type

Bug Report

### Component Name

ec2_tags, ec2_instance

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.21.2]
  config file =./ansible.cfg
  configured module search path = ['/Users/marcus/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/homebrew/Cellar/ansible/14.2.0_1/libexec/lib/python3.14/site-packages/ansible
  ansible collection location = /Users/marcus/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/homebrew/bin/ansible
  python version = 3.14.6 (main, Jun 10 2026, 10:03:53) [Clang 21.0.0 (clang-2100.0.123.102)] (/opt/homebrew/Cellar/ansible/14.2.0_1/libexec/bin/python)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

### Collection Versions

```console (paste below)
$ ansible-galaxy collection list
[WARNING]: Collection at '/opt/homebrew/Cellar/ansible/14.2.0_1/libexec/lib/python3.14/site-packages/ansible/_internal/ansible_collections/ansible/_protomatter' does not have a MANIFEST.json file, nor has it galaxy.yml: cannot detect version.                                                                                                              

# /Users/marcus/.ansible/collections/ansible_collections
Collection                               Version
---------------------------------------- -------
amazon.aws                               11.4.0 
ansible.mysql                            5.2.0  
ansible.netcommon                        8.6.2  
ansible.posix                            2.2.2  
ansible.utils                            6.1.0  
ansible.windows                          3.7.0  
arista.eos                               12.2.0 
[...]

Issue #3028, opened 2026-07-07
ec2 inventory plugin always throws 'tags' deprecation warning


dnmvisser:
### Summary

The ec2 inventory plugin seems to always emit a warning about the `tags` host variable. For example with this config:

```yaml
plugin: amazon.aws.aws_ec2
regions:
  - eu-central-1
```

The plugin works but I get a deprecation warning:

```console
(venv) root@trixie:~# ansible-inventory -i minimal.aws_ec2.yml --list >/dev/null
[DEPRECATION WARNING]: The 'tags' host variable is deprecated. Use 'ec2_tags' 
instead. This feature will be removed from amazon.aws in a release after 
2026-12-01. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
[WARNING]: Found variable using reserved name: tags
```

This is with:

```
(venv) root@trixie:~# ansible --version                 
ansible [core 2.18.18]
  config file = None
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/venv/lib/python3.13/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /root/venv/bin/ansible
  python version = 3.13.5 (main, May  5 2026, 21:05:52) [GCC 14.2.0] (/root/venv/bin/python3)
  jinja version = 3.1.6
  libyaml = True
```

```
(venv) root@trixie:~# ansible-config dump --only-changed
CONFIG_FILE() = None

GALAXY_SERVERS:
```

```
(venv) root@trixie:~# ansible-galaxy collection list

# /root/.ansible/collections/ansible_collections
Collection Version
---------- -------
amazon.aws 11.4.0 
```

```
(venv) root@trixie:~# pip list
Package         Version
--------------- -----------
ansible-core    2.18.18
boto3           1.43.41
botocore        1.43.41
cffi            2.1.0
cryptography    49.0.0
Jinja2          3.1.6
jmespath        1.1.0
MarkupSafe      3.0.3
packaging       26.2
pip             25.1.1
pycparser       3.0
python-dateutil 2.9.0.post0
PyYAML          6.0.3
resolvelib      1.0.1
s3transfer      0.19.0
six             1.17.0
urllib3         2.7.0
```

### Issue Type

Bug Report

### Component Name

ec2_inventory

### Ansible Version

```console (paste below)
[...]

PR #3085, opened 2026-09-09
feat(ec2_tag): add resource_tags alias for tags parameter

Romi1495:
##### SUMMARY

Adds `resource_tags` as an alias for the required `tags` parameter of the `ec2_tag` module, matching the convention already used by `ec2_instance`.

Refs #3063

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

ec2_tag

##### ADDITIONAL INFORMATION

The functional change is a single line in the argument spec:

```python
tags=dict(type="dict", required=True, aliases=["resource_tags"]),
```

`AnsibleModule` resolves aliases before populating `module.params`, so no downstream logic needed changing and `required=True` is still satisfied by either name.

Also included: documentation of the alias, an example using it, and three integration tests — set via the alias, idempotency, and removal via `tags` of a tag that was set via `resource_tags`. That last one is the meaningful guard: it only passes if both names resolve to the same parameter rather than two parallel paths.

`ec2_tag_info` was reviewed and deliberately left unchanged — it has no `tags` input parameter, only `resource`.

**On the reserved-name warning**

Issue #3063 also asks that using the alias stop ansible-core's `Found variable using reserved name 'tags'` warning. This PR does not do that, and I don't think an alias can.

`warn_if_reserved()` is called from exactly four places, all in `ansible/vars/manager.py`, and each passes variables or facts — never module parameter names. Verified with a module using this exact argspec:

| Scenario | Warning? |
|---|---|
| Module param literally named `tags` | No |
| Module param via `resource_tags` alias | No |
| Alias used, but a variable named `tags` exists | **Yes** |

The trigger is a variable named `tags`, not the module parameter. Where that variable comes from the `aws_ec2` inventory plugin, #3028 looks like the relevant issue. I've raised this on #3063 so refinement can decide whether to rescope that item or move it.

The alias still stands on its own merits as a consistency fix with `ec2_instance`.

**Testing**

`ansible-test sanity` passes on the changed files — 26 tests including validate-modules, ansible-doc, pylint, pep8, yamllint, and changelog. `compile`/`import` ran on Python 3.13 and 3.14 only, as older interpreters weren't available locally; CI will cover the rest. Integration tests were not run locally since they require AWS credentials.

Assisted-by: Claude Opus 5

### DIFF
--- a/changelogs/fragments/3063-ec2_tag-resource_tags-alias.yml
+++ b/changelogs/fragments/3063-ec2_tag-resource_tags-alias.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - ec2_tag - added ``resource_tags`` as an alias for the ``tags`` parameter, matching the convention already used by the ``ec2_instance`` module (https://github.com/ansible-collections/amazon.aws/issues/3063).
+  - ec2_tag - added ``resource_tags`` as an alias for the ``tags`` parameter, matching the convention already used by the ``ec2_instance`` module (https://github.com/ansible-collections/amazon.aws/pull/3085).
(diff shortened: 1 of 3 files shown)


oraNod (approved):
LGTM. Thanks @Romi1495

PR #3088, opened 2026-09-09
aws_ssm: fix Windows echo-disable hang and invalid-payload error

### LINKED ISSUE #3086: aws_ssm: disable_windows_echo_command() hangs on Windows targets (multi-line Add-Type sent inline over interactive session)
### Summary

`disable_windows_echo_command()` in `plugin_utils/ssm/terminalmanager.py` reliably hangs on a Windows target during `aws_ssm` connection setup. The command it sends is a multi-line `Add-Type @'...'@` here-string, written to the interactive SSM session via a single `stdin_write()` call containing real embedded newline bytes.

Confirmed via `-vvvv`, 100% reproducible on a real Windows Server 2019 target (`amazon.aws` 11.4.0, `ansible-core` 2.17.14): the session consistently stalls partway through the `Add-Type` block (around line 15 of ~20), and no further `WAIT_FOR_MATCH` output ever arrives. `ps -o pid,etime,time,stat` on the controller shows the process genuinely idle (near-zero CPU growth over minutes), not retrying.

### Root cause

Sending a large multi-line command with real embedded newlines to an interactive Windows PSSession over SSM appears to trigger PowerShell's `>>` continuation-prompt handling in a way the plugin's own terminal-matching logic doesn't expect, and the session never recovers. This is the exact class of problem `exec_command()`'s own S3-based execution path was built to avoid (see the code's own comment: *"The PTY echoes ALL stdin data to stdout before PowerShell can read it, causing timeouts on large commands"*, from #2909), but that fix was never applied to this bootstrap step — `disable_windows_echo_command()` still sends its command inline via `stdin_write()`.

### Fix

Rewrote the command to be a single physical line — using PowerShell backtick escapes (`` `n `` instead of real newline bytes, `` `" `` instead of bare double quotes) instead of the multi-line `@'...'@` here-string syntax. PowerShell still expands the escapes into equivalent multi-line C# source before handing it to `Add-Type`, but nothing containing an actual newline byte is sent over the interactive session anymore.

```python
# Before (plugin_utils/ssm/terminalmanager.py, disable_windows_echo_command):
command = (
    "Remove-Module PSReadLine -ErrorAction SilentlyContinue ; "
    "Add-Type -TypeDefinition @'\n"
    "using System;\n"
    "using System.Runtime.InteropServices;\n"
    "public class ConsoleHelper {\n"
    '    [DllImport("kernel32.dll", SetLastError = true)]\n'
    "    public static extern IntPtr GetStdHandle(int nStdHandle);\n"
    '    [DllImport("kernel32.dll", SetLastError = true)]\n'
    "    public static extern bool GetConsoleMode(IntPtr hConsoleHandle, out uint lpMode);\n"
    '    [DllImport("kernel32.dll", SetLastError = true)]\n'
    "    public static extern bool SetConsoleMode(IntPtr hConsoleHandle, uint dwMode);\n"
    "    public const int STD_INPUT_HANDLE = -10;\n"
    "    public const uint ENABLE_ECHO_INPUT = 0x0004;\n"
    "}\n"
    "'@ -ErrorAction SilentlyContinue ; "
    "$h = [ConsoleHelper]::GetStdHandle([ConsoleHelper]::STD_INPUT_HANDLE) ; "
    "$mode = [uint32]0 ; "
    "[void][ConsoleHelper]::GetConsoleMode($h, [ref]$mode) ; "
    "$mode = $mode -band (-bnot [ConsoleHelper]::ENABLE_ECHO_INPUT) ; "
    "[void][ConsoleHelper]::SetConsoleMode($h, $mode) ; "
    f"echo {end_mark}\n"
)

# After:
command = (
    "Remove-Module PSReadLine -ErrorAction SilentlyContinue ; "
    'Add-Type -TypeDefinition "using System;`n'
    "using System.Runtime.InteropServices;`n"
    "public class ConsoleHelper {`n"
    '    [DllImport(`"kernel32.dll`", SetLastError = true)]`n'
    "    public static extern IntPtr GetStdHandle(int nStdHandle);`n"
    '    [DllImport(`"kernel32.dll`", SetLastError = true)]`n'
    "    public static extern bool GetConsoleMode(IntPtr hConsoleHandle, out uint lpMode);`n"
    '    [DllImport(`"kernel32.dll`", SetLastError = true)]`n'
    "    public static extern bool SetConsoleMode(IntPtr hConsoleHandle, uint dwMode);`n"
    "    public const int STD_INPUT_HANDLE = -10;`n"
    "    public const uint ENABLE_ECHO_INPUT = 0x0004;`n"
    '}" -ErrorAction SilentlyContinue ; '
    "$h = [ConsoleHelper]::GetStdHandle([ConsoleHelper]::STD_INPUT_HANDLE) ; "
    "$mode = [uint32]0 ; "
    "[void][ConsoleHelper]::GetConsoleMode($h, [ref]$mode) ; "
    "$mode = $mode -band (-bnot [ConsoleHelper]::ENABLE_ECHO_INPUT) ; "
    "[void][ConsoleHelper]::SetConsoleMode($h, $mode) ; "
    f"echo {end_mark}\n"
)
```

The resulting command string contains exactly one real newline (the trailing Enter to submit it), down from 15+.

### Verification

Confirmed live against a real WorkSpaces-hosted Windows Server 2019 box over `aws_ssm`: `DISABLE WIN ECHO` now completes in ~1.8 seconds instead of hanging indefinitely at the same point every time.

### Environment

- `amazon.aws` 11.4.0
- `ansible-core` 2.17.14
- Target: Windows Server 2019 Datacenter (AWS WorkSpaces, hybrid SSM activation)
- Controller: macOS

### LINKED ISSUE #3087: aws_ssm: Windows module execution fails with "invalid payload" on the standard `type <file> | PowerShell -EncodedCommand` invocation shape
### Summary

Windows module execution over `aws_ssm` fails deterministically with `"invalid payload"` whenever ansible-core sends its standard `type <AnsiballZ file> | PowerShell -EncodedCommand <script>` invocation shape. This is likely the real root cause behind other "invalid payload" reports that looked intermittent (e.g. reported as such in our own testing before this was isolated) — it's actually 100% reproducible for this exact command shape, and only *looks* intermittent because `ansible-playbook` stops running further tasks on a host after the first task fails, so which task happens to hit this shape first varies run to run.

### Root cause

ansible-core's standard pattern for invoking a Windows module over any raw-command connection plugin is:

```
type C:\Windows\TEMP\...\AnsiballZ_win_ping.ps1 | PowerShell -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -EncodedCommand <base64>
```

This is a native OS pipe: `type` streams the AnsiballZ file's content (module code + a `\0\0\0\0`-delimited JSON payload) into PowerShell's stdin, and the base64-encoded script (`bootstrap_wrapper.ps1`, from `ansible-core` itself) reads it via the `$input` automatic variable:

```powershell
$exec_wrapper_str = $input | Out-String
$split_parts = $exec_wrapper_str.Split(@("`0`0`0`0"), 2, [StringSplitOptions]::RemoveEmptyEntries)
If (-not $split_parts.Length -eq 2) { throw "invalid payload" }
```

`WindowsCommandExecutor._decode_powershell_command()` in `plugin_utils/ssm/windows_executor.py` extracts the script to upload to S3 like this:

```python
encoded_match = re.search(r'-EncodedCommand\s+["\']?([A-Za-z0-9+/=]+)["\']?', cmd_str)
```

`re.search` matches `-EncodedCommand ...` anywhere in the string and the code only ever looks at `encoded_match.group(1)` — everything before the match, including the entire `type <file> |` prefix, is silently discarded. The decoded `bootstrap_wrapper.ps1` script is then uploaded to S3 alone and executed via `_generate_wrapper_without_stdin()`, which runs it standalone with `powershell -File $t` and **no piped input at all**.

Result: `$input` is empty inside `bootstrap_wrapper.ps1`, `$exec_wrapper_str` is empty, splitting on `\0\0\0\0` yields zero parts, and `$split_parts.Length -eq 2` fails — `"invalid payload"`, every single time a module invocation takes this exact command shape.

Confirmed via `-vvvv`: the failing `exec_command()` call always logs `EXEC: No stdin data` (i.e. `in_data=None`) right before this failure, because ansible-core relies entirely on the *native pipe in the command string itself* to supply the module payload for this invocation style, not on the connection plugin's own `in_data` parameter.

### Fix

Capture the piped prefix (if `-EncodedCommand` is preceded by anything ending in `|`) and reattach it to the wrapper's own invocation of the downloaded script, so the same native pipe still runs — just against the S3-downloaded copy of the script instead of the original inline command.

```python
def _decode_powershell_command(self, cmd_str: str) -> tuple[str, str]:
    script_to_upload = cmd_str
    pipe_prefix = ""
    encoded_match = re.search(r'-EncodedCommand\s+["\']?([A-Za-z0-9+/=]+)["\']?', cmd_str)

    if encoded_match:
        # Anything before -EncodedCommand that ends in a pipe is a native command
        # (e.g. `type <file> |`) whose stdout was meant to become this script's
        # $input. Preserve it so it can be reattached to the wrapper's own
        # invocation of the downloaded script.
        prefix_text = cmd_str[: encoded_match.start()]
        if "|" in prefix_text:
            pipe_prefix = prefix_text[: prefix_text.rfind("|") + 1].strip() + " "

        try:
            encoded_script = encoded_match.group(1)
            script_bytes = base64.b64decode(encoded_script)
            script_to_upload = script_bytes.decode("utf-16-le")
        except Exception:
            script_to_upload = cmd_str
            pipe_prefix = ""

    return script_to_upload, pipe_prefix
```

And in `_generate_wrapper_without_stdin()`, accept `pipe_prefix: str = ""` and use it right before the script invocation line:

```python
f"{pipe_prefix}powershell -NoProfile -NonInteractive -ExecutionPolicy Bypass -File $t ; "
```

(`execute()` is updated to unpack the tuple and pass `pipe_prefix` through accordingly.)

### Verification

Before the fix: 4/4 clean `win_ping` ad-hoc runs failed with `"invalid payload"`; a 3-task playbook stopped after task 2 hit the same error.

After the fix: 5/5 `win_ping` ad-hoc runs succeeded (`rc=0`), and the same 3-task playbook completed cleanly (`ok=3 changed=0 unreachable=0 failed=0`). All runs confirmed via `-vvv`/`-vvvv` that they hit the exact `type <file> | PowerShell -EncodedCommand` shape and still succeeded.

### Environment

- `amazon.aws` 11.4.0
- `ansible-core` 2.17.14
- Target: Windows Server 2019 Datacenter (AWS WorkSpaces, hybrid SSM activation)
- Controller: macOS

### Related

This is the actual root cause behind the "invalid payload" symptom we'd previously only been able to characterize as intermittent/timing-dependent in unrelated internal testing on a different Windows box, before this specific command shape was isolated here.

treed26:
### SUMMARY

Fixes two distinct bugs in the `aws_ssm` connection plugin's Windows execution path, both found and verified live against a real Windows Server 2019 target reached via AWS SSM (WorkSpaces hybrid activation).

Fixes #3086
Fixes #3087

### ISSUE TYPE

- Bugfix Pull Request

### COMPONENT NAME

aws_ssm connection plugin

### ADDITIONAL INFORMATION

**Bug 1 (#3086):** `disable_windows_echo_command()` sent a multi-line `Add-Type @'...'@` here-string as a single interactive-session write, containing real embedded newline bytes. This reliably hung partway through the block on a Windows target (confirmed via `-vvvv`, 100% reproducible, always stalling at the same point). Fixed by rewriting the command as a single physical line using PowerShell backtick escapes (`` `n ``, `` `" ``) instead of a real multi-line here-string. PowerShell still expands the escapes into equivalent multi-line C# source before handing it to `Add-Type`, but nothing containing an actual newline byte is sent over the interactive session anymore.

**Bug 2 (#3087):** ansible-core's standard Windows module-invocation shape is `type <AnsiballZ file> | PowerShell -EncodedCommand <script>` — a native OS pipe feeding the file's `\0\0\0\0`-delimited payload into the script's `$input`. `_decode_powershell_command()`'s regex only located `-EncodedCommand` and everything after it, silently discarding the `type <file> |` prefix. The decoded script then ran standalone via `_generate_wrapper_without_stdin()` with no piped input at all, so `$input` was empty and ansible-core's own `bootstrap_wrapper.ps1` sanity check (`$split_parts.Length -eq 2`) threw `"invalid payload"` — deterministically, every time this exact command shape occurred. Fixed by capturing the piped prefix and reattaching it to the wrapper's own invocation of the downloaded script.

**Verification:** before these fixes, 4/4 ad-hoc `win_ping` runs failed with `"invalid payload"`, and a 3-task playbook stopped after task 2 hit the same error. After both fixes: 5/5 ad-hoc `win_ping` runs succeeded (`rc=0`), and the same 3-task playbook completed cleanly (`ok=3 changed=0 unreachable=0 failed=0`). 8/8 clean executions total, all confirmed (via `-vvv`/`-vvvv`) to have hit the exact previously-broken command shapes.

Environment: `amazon.aws` 11.4.0, `ansible-core` 2.17.14, target Windows Server 2019 Datacenter (AWS WorkSpaces, hybrid SSM activation), macOS controller.


treed26:
Closing this PR — withdrawing the code contribution itself. The two underlying bugs (#3086, #3087) are still real and documented there for anyone else who hits them; we're just no longer pursuing the upstream fix through this PR, and are instead applying an equivalent local patch scoped entirely to our own repo.

PR #3069, opened 2026-08-19
Fix/aws rds refactor

Romi1495:
##### SUMMARY

  Refactor `rds_cluster_param_group` module to align error handling with patterns established in #2119 and
  #2138. Part of https://github.com/ansible-collections/amazon.aws/issues/2003.

##### ISSUE TYPE
Refactoring Pull Request

##### COMPONENT NAME

`rds_cluster_param_group`
`module_utils/_rds/api.py`
`module_utils/_rds/common.py`

##### ADDITIONAL INFORMATION

  module_utils/_rds/common.py:
  - Add DBParameterGroupNotFound to RDSErrorHandler._is_missing() error code list

  module_utils/_rds/api.py:
  - Refactor `describe_db_cluster_parameter_groups()` to use `@RDSErrorHandler.list_error_handler` decorator
  instead of manual `try/except` with `is_boto3_error_code`
  - Refactor `describe_db_cluster_parameters()` to use `@RDSErrorHandler.list_error_handler decorator`
  instead of manual `try/except `with i`s_boto3_error_code`

  rds_cluster_param_group module:
  - Add PEP 257 docstrings to `modify_parameters()`, `ensure_present()`, and `ensure_absent()` functions

Assisted by Claude opus 4.6

PR #3089, opened 2026-09-10
refactor(rds_cluster_param_group): use shared module_utils/rds utilities

mandar242:
##### SUMMARY

Refactors `rds_cluster_param_group` onto the shared `module_utils/rds` utilities, continuing the migration started in #3069 and following the patterns established by #2119 (shared boto3 client helpers and error handling) and #2138 (shared `describe_*` functions).

This is intended to be a pure refactor, no behaviour change.

**`module_utils/_rds/api.py`**

- `describe_db_cluster_parameter_groups()` and `describe_db_cluster_parameters()` now take `(client, **params)` like every other `describe_*` helper in the module. Both previously took a `module` argument that became completely unused once #3069 replaced their manual `try/except` blocks with the `@RDSErrorHandler.list_error_handler` decorator. Filter construction moves to the callers, which is where the module params live.
- Renamed the `connection` argument to `client` in the cluster parameter group create/delete/modify helpers, for consistency with the rest of the file.

**`rds_cluster_param_group`**

- Dropped the `botocore` import and the `try/except` around `module.client()`. `AnsibleAWSModule` already handles connection failures, so this was dead code. The client-level `retry_decorator` is kept, matching #3072.
- `AnsibleRDSError` raised by the shared helpers is now caught once at the module boundary in `main()`.
- Split `ensure_present()` into `get_parameter_group()`, `create_parameter_group()` and `update_parameter_group()`, with type hints and Google-style docstrings throughout.
- Reuse the parameter group returned by the create call instead of re-describing it, removing a redundant `describe_db_cluster_parameter_groups()` call. `CreateDBClusterParameterGroup` and `DescribeDBClusterParameterGroups` return the same `DBClusterParameterGroup` shape in the botocore service model, so the module output is unchanged.
- Index the current parameters by name when comparing them against the requested ones, instead of scanning the full list once per requested parameter.

**`rds_cluster_param_group_info`**

- Adopted the new describe helper signatures and dropped the same redundant `botocore` import and connection `try/except`.

##### ISSUE TYPE
- Refactoring Pull Request

##### COMPONENT NAME

rds_cluster_param_group
rds_cluster_param_group_info
module_utils/_rds/api.py

##### ADDITIONAL INFORMATION

**Behaviour**

Everything is intended to be behaviour-preserving: the returned `db_cluster_parameter_group` keys, the parameter validation failure messages (`Could not find parameter with name: ...` and `The parameter ... cannot be modified`, still raised via `module.fail_json()` so they keep their exact text), the set of parameters sent to `ModifyDBClusterParameterGroup`, the `tags: {}` no-op, the check mode early exit and its message, idempotency, and the "does not exist" message on `state=absent` are all unchanged.

The only externally observable difference is one fewer `DescribeDBClusterParameterGroups` call on the create path.

**Testing**

- New unit tests: `tests/unit/plugins/modules/test_rds_cluster_param_group.py` (21 tests) and `tests/unit/plugins/modules/test_rds_cluster_param_group_info.py` (4 tests), plus 10 tests in `test_rds_api.py` covering the shared cluster parameter group helpers, which had no unit coverage before.
- Full RDS unit suite passes (852 tests).
- `ansible-test sanity` passes on all changed files.
- Existing integration tests under `tests/integration/targets/rds_cluster_param_group` are unchanged and each assertion was traced against the new code paths, but they have not been run against live AWS in this branch. A CI run would be good confirmation.

Assisted-by: Claude Opus 5


beeankha (approved):

oraNod (approved):
LGTM. Thanks @mandar242

PR #3092, opened 2026-09-11
event_query: 13 of 15 audit queries report no node at all -- 2 raise a jq error that discards the whole event, and status/tags in the dedup key re-count every resource

fistilis:
##### SUMMARY

`extensions/audit/event_query.yml` is executable code that runs inside the Automation Platform controller (`awx/main/tasks/host_indirect.py`) and feeds subscription node counting. Nothing in this repository executes it against a real module result, so it has shipped green regardless of what it does.

Running each query against the shape its module actually returns, **only 2 of the 15 produce a countable node**:

```
before                          after
  COUNTED       2                 COUNTED      15
  DROPPED      11
  EVENT ERROR   2
```

<details>
<summary>Full before / after, per module</summary>

```
================ BEFORE
ec2_instance_info         COUNTED      {"id": "i-0abc...", "status": "running", "tags": {...}}
ec2_vpc_igw_info          EVENT ERROR  jq: error: Cannot iterate over null (null)
ec2_vpc_nat_gateway_info  DROPPED      no non-null top-level `name`
ec2_vpc_net_info          COUNTED      {"id": "vpc-0123...", "status": "available", "tags": {...}}
ec2_vpc_peering_info      DROPPED      no non-null top-level `name`
ec2_vpc_route_table_info  DROPPED      no non-null top-level `name`
ec2_vpc_subnet_info       DROPPED      no non-null top-level `name`
ec2_vpc_vgw_info          DROPPED      no non-null top-level `name`
ec2_vpc_vpn_info          DROPPED      no non-null top-level `name`
elb_application_lb_info   DROPPED      no non-null top-level `name`
elb_classic_lb_info       DROPPED      no non-null top-level `name`
rds_cluster_info          DROPPED      no non-null top-level `name`
rds_instance_info         DROPPED      no non-null top-level `name`
s3_bucket_info            DROPPED      no non-null top-level `name`
s3_object_info            EVENT ERROR  jq: error: Cannot iterate over null (null)

================ AFTER
ec2_instance_info         COUNTED      {"instance_id": "i-0abc1234def567890"}
ec2_vpc_igw_info          COUNTED      {"internet_gateway_id": "igw-0123456789abcdef0"}
ec2_vpc_nat_gateway_info  COUNTED      {"nat_gateway_id": "nat-0123456789abcdef0"}
ec2_vpc_net_info          COUNTED      {"vpc_id": "vpc-0123456789abcdef0"}
ec2_vpc_peering_info      COUNTED      {"vpc_peering_connection_id": "pcx-0123456789abcdef0"}
ec2_vpc_route_table_info  COUNTED      {"route_table_id": "rtb-0123456789abcdef0"}
ec2_vpc_subnet_info       COUNTED      {"subnet_id": "subnet-0123456789abcdef0"}
ec2_vpc_vgw_info          COUNTED      {"vpn_gateway_id": "vgw-0123456789abcdef0"}
ec2_vpc_vpn_info          COUNTED      {"vpn_connection_id": "vpn-0123456789abcdef0"}
elb_application_lb_info   COUNTED      {"load_balancer_arn": "arn:aws:elasticloadbalancing:..."}
elb_classic_lb_info       COUNTED      {"load_balancer_name": "legacy-elb"}
rds_cluster_info          COUNTED      {"db_cluster_resource_id": "cluster-ABCDEFGHIJKLMNOP"}
rds_instance_info         COUNTED      {"dbi_resource_id": "db-ABCDEFGHIJKLMNOP"}
s3_bucket_info            COUNTED      {"bucket_name": "example-prod-bucket"}
s3_object_info            COUNTED      {"object_name": "data/a.json"}   (with object_details set)
```
</details>

##### Four independent defects

**1. No top-level `name` — 13 queries.** `host_indirect.py` does `if name is None: continue`. Thirteen queries emit no `name` key at all, so every load balancer, subnet, NAT gateway, internet gateway, virtual gateway, route table, VPN connection, VPC peering, S3 bucket, S3 object, RDS instance and RDS cluster is discarded on arrival. Only `ec2_instance_info` and `ec2_vpc_net_info` emit one.

**2. A jq error takes the whole event — 2 queries.** A jq error is caught one level up, at the event, so it discards the audit records of *every other module in that event* too, not just its own.

* `.object_info[]` — `s3_object_info` returns `object_info` only when `object_details` is set. Called with `bucket_name` alone, which is its most common form, it returns `s3_keys` and the query raises `Cannot iterate over null`. Called with `object_name` but no `object_details` it returns `object_data` only, with no `object_name` key anywhere, so `canonical_facts.name` is null and the record is dropped for reason 3 instead.
* `.attachments | map(...)` — a detached internet gateway has no `attachments` key, which is the normal state of a freshly created gateway.

Every source array is now `[]?`-guarded and type-checked before it is indexed.

**3. Volatile fields in the dedup key.** `canonical_facts` is the sole dedup key — `results[hashable_facts]` — so anything in it that changes re-counts the same resource as a new one. `status` and `tags` were in all fifteen:

```
ONE EC2 instance (i-0abc1234def567890) seen running, stopped, and re-tagged
  -> distinct audit rows: 3 (expected 1)
     {"id": "i-...", "status": "running", "tags": [["Name","web-01"],["env","prod"]]}
     {"id": "i-...", "status": "running", "tags": [["Name","web-01"],["env","prod"],["owner","team-a"]]}
     {"id": "i-...", "status": "stopped", "tags": [["Name","web-01"],["env","prod"]]}
```

`rds_instance_info` and `rds_cluster_info` additionally carried the user-supplied identifier alongside the immutable resource id, so a rename counted the database twice. `elb_application_lb_info` was keyed on the load balancer *name*, which AWS allows to be reused after deletion. All of these moved to `facts`, which is not hashed.

**4. Taxonomy not normalized — all 15.** Values must be `lowercase_with_underscores` (`^[a-z0-9]+(_[a-z0-9]+)*$`). `PublicCloud` and `public_cloud` are two separate buckets downstream, and `NAT Gateway`, `Route table`, `Internet Gateway`, `Virtual Gateway`, `VPN Connection` and `VPC Peering` match no bucket at all.

One judgement call worth flagging: the two ambiguous RDS values `Instance` and `Cluster` became `database_instance` and `database_cluster` rather than `instance`/`cluster`, because those collide with unrelated resource kinds in a cross-collection rollup. Happy to change them to the literal normalization if you would rather keep the mapping one-to-one.

##### Why no test caught any of this

The `node_query_*` integration targets assert equality against expected results that were derived from the queries' own output — `templates/*.j2` literally reproduce the query logic and then compare it to the query. An assertion captured from output agrees with that output by construction: it can detect drift, never wrongness. They are also never selected on a change that touches only `extensions/audit/`, because the test splitter maps changed files to targets by module path.

This PR adds `tests/unit/extensions/audit/test_event_query_contract.py`, which asserts the contract the *consumer* imposes rather than the output the query happens to produce. It is static, needs no AWS credentials and no live resources, so it runs in the units job on every pull request including query-file-only ones. It fails on `main` and passes here.

The integration targets are updated to the corrected contract as well. Every one of their new expected results was checked locally against the queries by rendering the templates against the real module return shapes.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
extensions/audit/event_query.yml

##### ADDITIONAL INFORMATION

Same class of defect as ansible-collections/vmware.vmware#414, ansible-collections/google.cloud#781, ansible-collections/azure#2365 and CiscoDevNet/intersight-ansible#354 — nine collections ship a query file and none of them execute it in CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelogs/fragments/event_query-contract-tests.yml
+++ b/changelogs/fragments/event_query-contract-tests.yml
@@ -1,0 +1,5 @@
+---
+trivial:
+  - tests - add unit tests asserting the ``extensions/audit/event_query.yml`` output contract
+    (non-null top-level ``name``, complete taxonomy, normalized taxonomy values) and correct the
+    ``event_query_parser`` return sample to match that contract.
--- a/changelogs/fragments/event_query-contract-tests.yml
+++ b/changelogs/fragments/event_query-contract-tests.yml
@@ -1,5 +1,30 @@
 ---
+bugfixes:
+  - event_query - emit a non-null top-level ``name`` from every query. Thirteen of the fifteen
+    queries emitted none at all, and the indirect node counting consumer discards any record
+    without one (https://github.com/ansible-collections/amazon.aws/pull/3092).
+  - event_query - guard the source array of every query with ``[]?``. ``.object_info[]`` and
+    ``.internet_gateways[] | .attachments | map(...)`` raised "Cannot iterate over null" for
+    result shapes both modules routinely return, and a jq error is handled at the event, so it
+    discarded the audit records of every other module in the same event
+    (https://github.com/ansible-collections/amazon.aws/pull/3092).
+  - event_query - move ``status``, ``tags`` and the user-supplied RDS identifiers out of
+    ``canonical_facts`` and into ``facts``. ``canonical_facts`` is the sole deduplication key, so
+    stopping an instance, editing a tag or renaming a database counted the resource again
+    (https://github.com/ansible-collections/amazon.aws/pull/3092).
+  - event_query - key ``elb_application_lb_info`` on the load balancer ARN rather than its name,
+    which can be reused after the load balancer is deleted
+    (https://github.com/ansible-collections/amazon.aws/pull/3092).
+minor_changes:
+  - event_query - normalize every ``infra_type``, ``infra_bucket`` and ``device_type`` value to
+    ``lowercase_with_underscores``. ``PublicCloud`` and ``public_cloud`` are two separate buckets
+    downstream, and ``NAT Gateway`` and ``Route table`` matched no bucket at all. The two
+    ambiguous RDS values are now ``database_instance`` and ``database_cluster``
+    (https://github.com/ansible-collections/amazon.aws/pull/3092).
 trivial:
   - tests - add unit tests asserting the ``extensions/audit/event_query.yml`` output contract
-    (non-null top-level ``name``, complete taxonomy, normalized taxonomy values) and correct the
-    ``event_query_parser`` return sample to match that contract.
+    (non-null top-level ``name``, non-null and identity-only ``canonical_facts``, complete and
+    normalized taxonomy) and correct the ``event_query_parser`` return sample to match it.
+  - tests - update the ``node_query_*`` integration targets to the corrected contract. Their
+    expected results were derived from the queries' own output, so they agreed with every defect
+    above by construction.
--- a/extensions/audit/event_query.yml
+++ b/extensions/audit/event_query.yml
@@ -1,178 +1,247 @@
 ---
+# Indirect node counting queries for the amazon.aws collection.
+#
+# These jq expressions run inside the Automation Platform controller
+# (awx/main/tasks/host_indirect.py) against each module's result, and their
+# output feeds subscription node counting. Nothing in this repository executes
+# them against a real module result, so
+# tests/unit/extensions/audit/test_event_query_contract.py asserts the contract
+# the controller imposes on the output.
+#
+# THE CONTRACT
+# ------------
+# name             must be present and never null. `if name is None: continue`
+#                  -- a record with no top-level `name` key is discarded.
+# canonical_facts  must be present, truthy, and free of null at ANY depth.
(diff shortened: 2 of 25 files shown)


fistilis:
Marking this ready for review.

@alinabuzachis — this supersedes #3066, which I have closed. #3066 fixed the missing top-level `name`; this PR is a strict superset of it and also fixes the three other defect classes in the same file (jq errors that discard the whole event, volatile fields in the dedup key, and unnormalized taxonomy), so reviewing one diff rather than two should be less work overall. Your two review points on #3066 are carried over here: the README section on the top-level `name`, and the consistent `.name // .tags.Name // <id> // "UNKNOWN"` identifier fallback.

Worth flagging separately: this repository's `all_green_check.yml` has completed with **zero jobs executed** on every run since 2026-08-06, on PRs and on `main` alike — re-verified today. Sanity and unit tests have not actually run for over five weeks, which is why the defects here shipped unnoticed, and it means a green tick on this PR does not yet mean the new tests passed.

PR #3097, opened 2026-09-18
feat(cloudformation): remove deprecated template parameter

mandar242:
##### SUMMARY

The template parameter was deprecated in 7.6.0 and the removal date of 2026-05-01 has passed. Use template_body with the file or template lookup instead.

BREAKING CHANGE: the template parameter is no longer supported in cloudformation.

Assisted-by: Cursor Composer

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

cloudformation


plugins/modules/cloudformation.py
```diff
@@ -244,7 +228,7 @@
     stack_name: "ansible-cloudformation"
     state: "present"
     region: "us-east-1"
-    template: "files/cloudformation-example.json"
+    template_body: "{{ lookup('file', 'cloudformation-example.json') }}"
```
abikouo, line 231:
```suggestion
    template_body: "{{ lookup('ansible.builtin.file', 'cloudformation-example.json') }}"
```
Can we do the same for all occurrences of the `file` lookup?

abikouo (changes requested):
Looks good overall, just a minor remark regarding the module example block

oraNod (approved):
Agree with the FQCN comment from @abikouo but rest seems clean and complete. Thanks @mandar242

alinabuzachis (approved):
Agreed on Aubin's suggestion, otherwise, LGTM!

